PR #1, opened 2023-12-05
Optimize CnaEvent lookup and storage in CnaUtil

sgicbpc:
This commit improves the performance of the storeCnaEvents method by replacing the stream-based lookup with a HashSet contains check. This change reduces the computational complexity from O(n) to O(1) for the lookup operation, resulting in a significant performance gain when processing large collections of CnaEvent objects.

Note: This PR is carried over from https://github.com/cBioPortal/cbioportal/pull/10478 

Requesting review and merge from @JREastonMarks . Thank you!


JREastonMarks:
Hi,
We haven't forgotten about this PR either. I plan on merging it in after the 6.0.0 release which will hopefully happen soon.
~Jeremy

PR #2, opened 2024-01-11
remove ssl_mode parameter if not enabled

nr23730:
Currently the pararmeter `connection_kwargs['ssl_mode'] = 'DISABLED'` is set, even if ssl_mode is completely off. This is on one side not necessary and on the other side do many distributions nowadays ship with mariadb instead of mysql (and use mysql just as alias for mariadb). The latter one does not support the `ssl_mode`.

Therefore I propose to remove that parameter if it is disabled.


JREastonMarks:
Hi,
We haven't forgotten about this PR either. I plan on merging it in after the 6.0.0 release which will hopefully happen soon.
~Jeremy

JREastonMarks (approved):

JREastonMarks:
lgtm

PR #3, opened 2024-01-11
migrate to spring db naming scheme

nr23730:
As 6.0.0 migrated to spring db I propose to also change this in the core library. Then the db settings in the application.properties only need to be defined once.

Goes together with: https://github.com/cBioPortal/cbioportal/pull/10537


inodb (approved):
Thanks so much @nr23730 ! Great catch!

nr23730:
You're welcome! However, this need's some more love, as the java side won't connect to the db right now.

JREastonMarks:
Hi,
I wanted to let you know we haven't forgotten about this PR. I plan on merging it in after the 6.0.0 release which will hopefully happen soon.
~Jeremy

JREastonMarks (approved):
lgtm

PR #6, opened 2024-02-05
Clarify what this repo represents

inodb:
Still some confusion about this being called legacy, so made some changes in phrasing


JREastonMarks (approved):

sheridancbio (approved):
👍 

If the entire cbioportal-core repository is intended to become deprecated code then maybe we don't need to worry about stripping away any of the other services in cbioportal-core (such as the webservice.do functionality). But if there are dependencies which we think deserve continued support after deprecation we might try to identify those now so we can give better guidance to the community on what portions of this repo will be eventually dropped v. which will continue to be supported.

inodb:
@sheridancbio That's a good point re describing exactly what we intend to become deprecated. I'm not entirely sure

At the moment `webservice.do` is no longer working, so if no one complaints that would be one of the things that'll be deprecated

PR #7, opened 2024-02-06
Unify validation of IDs

oplantalech:
The validation of IDs is scattered and it is even different between different ID types (Patient ID, Generic Assay ID...). This PR unifies the validation of IDs into a single variable.


JREastonMarks (approved):
LGTM. I like that simplifies the validation process

PR #10, opened 2024-02-14
Ignore gene described in the CNA Discrete Long file if it cannot be found in the database

### LINKED ISSUE #8: Loader crashes when an unknown gene is present in a Discrete CNA Long file
There is a bug in the loader of the Discrete CNA Long file: when a gene not found in the database is placed in this file, the loader attempts to load it and it crashes - instead of skipping it (as reported by the validator).

oplantalech:
Fix #8


JREastonMarks (approved):
lgtm, thanks for adding this!

PR #13, opened 2024-02-26
Add java_opts to cbioportalImporter.py

oplantalech:
Currently we can only pass the `jar_path` to `cbioportalImporter.py`. This adds `java_opts` as a new option to `cbioportalImporter.py` to allow passing any `JAVA_OPTS` parameter. This allows us pass max memory usage etc. The `jar_path` option is retained in a backward-compatible manner

### DIFF
--- a/src/main/resources/scripts/importer/cbioportalImporter.py
+++ b/src/main/resources/scripts/importer/cbioportalImporter.py
@@ -444,10 +444,9 @@ def check_dir(study_directory):
 def add_parser_args(parser):
     parser.add_argument('-s', '--study_directory', type=str, required=False,
                         help='Path to Study Directory')
-    parser.add_argument('-jvo', '--java_opts', type=str, required=False,
+    parser.add_argument('-jvo', '--java_opts', type=str, default=os.environ.get('JAVA_OPTS'),
                         help='Path to specify JAVA_OPTS for the importer. \
-                        (default: locates the jar path relative to the import script \
-                        and passes it as the JAVA_OPTS)')
+                        (default: gets the JAVA_OPTS from the environment)')
     parser.add_argument('-jar', '--jar_path', type=str, required=False,
                         help='Path to scripts JAR file')
     parser.add_argument('-meta', '--meta_filename', type=str, required=False,
--- a/src/main/resources/scripts/importer/cbioportalImporter.py
+++ b/src/main/resources/scripts/importer/cbioportalImporter.py
@@ -444,8 +444,13 @@ def check_dir(study_directory):
 def add_parser_args(parser):
     parser.add_argument('-s', '--study_directory', type=str, required=False,
                         help='Path to Study Directory')
+    parser.add_argument('-jvo', '--java_opts', type=str, required=False,
+                        help='Path to specify JAVA_OPTS for the importer. \
+                        (default: locates the jar path relative to the import script \
+                        and passes it as the JAVA_OPTS)')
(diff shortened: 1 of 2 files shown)


src/main/resources/scripts/importer/cbioportalImporter.py
```diff
@@ -520,18 +525,25 @@ def main(args):
...
         except FileNotFoundError as e:
             print(e)
             sys.exit(2)
-        print('Data loading step using', args.jar_path)
+        print('Data loading step using', jar_path)
         print()
+        if args.java_opts is None:
+            args.java_opts = f"-cp {jar_path}"
+        else:
+            args.java_opts = f"-cp {jar_path} {args.java_opts}"
```
MatthijsPon, line 543: Will this not become `-cp -cp JAR_PATH OTHER_JAVA_OPTS` if -cp is specified in java_opts?

src/main/resources/scripts/importer/cbioportalImporter.py
```diff
@@ -444,8 +444,13 @@ def check_dir(study_directory):
 def add_parser_args(parser):
     parser.add_argument('-s', '--study_directory', type=str, required=False,
                         help='Path to Study Directory')
+    parser.add_argument('-jvo', '--java_opts', type=str, required=False,
+                        help='Path to specify JAVA_OPTS for the importer. \
+                        (default: locates the jar path relative to the import script \
+                        and passes it as the JAVA_OPTS)')
     parser.add_argument('-jar', '--jar_path', type=str, required=False,
-                        help='Path to scripts JAR file')
+                        help='DEPRECATED ARGUMENT. Please, use -jvo/--java_opts instead. If you \
```
MatthijsPon, line 452: Why not:
```
if args.jar_path:
    args.java_opts = f"-cp {args.jar_path} {args.java_opts}"
```
Then there is no 'need' to deprecate and you can replace it with something like 'WILL BE DEPRECATED IN THE FUTURE'?

MatthijsPon (commented):
Looks good, would be nice to be able to throttle memory for the loader, some questions:

src/main/resources/scripts/importer/cbioportalImporter.py
oplantalech: Hmm... but we want people to move away from using `jar_path`, right? We should then force them somehow to use the new attribute

JREastonMarks (changes requested):
I like the idea of adding java_opts as a parameter to these scripts. However, I agree with @MatthijsPon that this should be a warning at the most and we should continue to support jar_path. Asking all sites that use this to change to a new way of launching this is a big ask especially with no warning. We are doing that enough with spinning out CORE into it's own repo.

inodb:
Since `JAVA_OPTS` is an env parameter, another way of going about it is to pass the environment from the main process:

https://stackoverflow.com/questions/2231227/python-subprocess-popen-with-a-modified-environment

Avoids adding additional code to handle another cli parameter and u could e.g. set java opts directly as env on the container

oplantalech:
@inodb I've tried to set the `JAVA_OPTS` from the environment as the default for `--java_opts` (so only if `--java_opts` is not specified). Not sure if that's what we want, though. Also, if now the `JAVA_OPTS` from the environment defines the classpath, then we use that classpath instead of locating the JAR file relative to the import script. Is that what we want?

inodb (approved):
LGTM - thank you!

PR #16, opened 2024-03-01
Dockerise

forus:
Changes:

- Moved scripts from the jar artifact to the project's root directory, improving the development experience by making the environment more familiar and convenient for script developers.

- Imported the `requirements.txt` file for installing Python dependencies from the cBioPortal project, ensuring consistency and ease of setup for Python-related tasks.

- Developed a new Docker image that is fully equipped with all necessary dependencies, offering a ready-to-use environment for running the Java loader or executing scripts.

Follow up items in: https://github.com/cBioPortal/cbioportal/issues/10734


inodb:
@forus is the intention here to eventually have a separate docker image just for importing?

Dockerfile
```diff
@@ -0,0 +1,52 @@
+# Stage 1: Build the application
+FROM maven:3.9.6-amazoncorretto-21 as builder
```
haynescd, line 2: Opinion, but i would stick to the same image as cbioportal 
maven:3-eclipse-temurin-21

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+WORKDIR /app
+
+# Copy the java source files and the pom.xml file into the image
+COPY src ./src
+COPY pom.xml .
+
+# Build the application
+RUN mvn clean package
+
+FROM python:3.11
```
haynescd, line 14: Also just to point out (not wrong) I would just think its easier to start with a java 21 image and install python... Instead of the other way around.

Cbioportal DockerFile
https://github.com/cBioPortal/cbioportal/blob/14e00a6c5580700b96152c45f34baf150e652c89/docker/web-and-data/Dockerfile#L29

Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+
+# Set the working directory in the Maven image
+WORKDIR /app
+
+# Copy the java source files and the pom.xml file into the image
+COPY src ./src
+COPY pom.xml .
+
+# Build the application
+RUN mvn clean package
```
haynescd, line 12: I would -DskipTests

.gitignore
```diff
@@ -23,6 +23,8 @@
 hs_err_pid*
 replay_pid*
 /target/
+# the final jar artifact is placed in the root of the project
+/core-*.jar
```
haynescd, line 27: this isn't needed 
*.jar captures that

Dockerfile
haynescd: Another reason I think it would be easier to use base image eclipse-temurin:21 is you do not have to worry about having a exact URL

requirements.txt
```diff
@@ -0,0 +1,6 @@
+requests==2.20.0
```
haynescd, line 1: Probably resync this is already in the repo

haynescd (changes requested):

forus:
> @forus is the intention here to eventually have a separate docker image just for importing?

@inodb yes, exactly. This PR creates a separate docker image just for importing. What's left is to update the docs and clean up the `cbioportal` docker file from pulling `cbioportal-core` jar.

Dockerfile
forus: yes, it makes sense. I made it this way as I thought we plan to get rid of java loader at some point. But from having couple of calls, I've gotten impression that's the java loader is here to stay for a while.

Dockerfile
forus: Why? Keep in mind that we leave out integration tests with this command and run only unit tests when https://github.com/cBioPortal/cbioportal-core/pull/29 is in.

requirements.txt
forus: Do you mean that this dependency is transitive and is brought by another one? We don't have `resync` package in here, right?

forus:
> > @forus is the intention here to eventually have a separate docker image just for importing?
> 
> @inodb yes, exactly. This PR creates a separate docker image just for importing. What's left is to update the docs and clean up the `cbioportal` docker file from pulling `cbioportal-core` jar.

@inodb but those actions in the TODO should happen now, but later after some deprecation time. Does it make sense?

inodb:
@ruslan-forostianov that makes sense. I created separate follow-up items for these fixes

haynescd (approved):

haynescd (changes requested):
Everything looks good except the python gh action

forus:
> Everything looks good except the python gh action

@haynescd thanks for catching it. I've pushed the fix. Please have another look.

haynescd (approved):
lgtm

PR #24, opened 2024-03-14
Fix core dep issues and cna utils

haynescd:


sheridancbio:
This PR has reverted #1

PR #25, opened 2024-03-15
Speed up CnaEvent lookup during import

sheridancbio:
- store existing events from the database in a HashMap instead of HashSet
- retrieval from HashMap uses CnaEvent.Event equals() and hashMap() semantics
- retrieval is neccessary in order to obtain the associated event_id from the db record
- this avoids a linear search through the set of all CnaEvent.Events in the database

### DIFF
--- a/src/main/java/org/mskcc/cbio/portal/scripts/ImportCnaDiscreteLongData.java
+++ b/src/main/java/org/mskcc/cbio/portal/scripts/ImportCnaDiscreteLongData.java
@@ -45,7 +45,7 @@ public class ImportCnaDiscreteLongData {
     private String genePanel;
     private final DaoGeneOptimized daoGene;
     private CnaUtil cnaUtil;
-    private Set<CnaEvent.Event> existingCnaEvents = new HashSet<>();
+    private Map<CnaEvent.Event, CnaEvent.Event> existingCnaEvents = new HashMap<>();
     private int samplesSkipped = 0;
     private Set<String> namespaces;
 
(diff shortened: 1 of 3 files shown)


sheridancbio:
This is an in-progress attempt to implement an efficient lookup of existing CnaEvent.Event references as intended by #1 
The prior PR had a bug, in that it obtained the event_id value from the newly created CnaEvent as parsed from the data_cna.txt file. However, the constructed CnaEvent objects coming from that parsing will not have any event_id set (they will default to 0). Instead, the event_id must be retrieved from an existing CnaEvent.Event object retrieved from the database through the DaoCnaEvent.getAllCnaEvents() call.

Instead of storing the CnaEvent.Events in a java Set, this PR stores them in a Map, mapping each event to itself. This allows retrieval from the Hashmap using .get(), which is efficient (more efficient than the linear search). In order for that to work correctly, the equals() function and hashCode() function of CnaEvent.Event has been overridden (previously) to function based only on gene and alteration fields (not event_id). The previous Map was demoted to a Set in this PR: https://github.com/cBioPortal/cbioportal/pull/9847
This PR reverses that and restores the previous HashMap lookup.

sheridancbio:
Note : the validity of using the HashMap representation to retrieve the event_id populated objects based on a non-event_id argument to the Map.get(key) function was tested. However, a built importer with these code changes failed to obtain the event_id values as expected. Some additional debugging is needed ... probably related to the equals() and hashCode() functions of the other types contained inside of the CnaEvent.Event type.

PR #27, opened 2024-03-18
Change Assert() into Warning for db.user/db.password/...

sheridancbio:
/- validate all 4 required data source parameters are present (including url)
	- use error logging and exception throwing instead of assert()
- instead of exiting with error, log a warning if deprecated properties are set
       - (the deprectaed properties are currently needed by the msk importer)

---------

This change is needed because the msk importer currently requires the deprecated parameters:
db.user, db.password, db.connection_string, db.portal_db_name, db.host, db.driver

The change of the Assert() statement into a logged warning in JdbcDataSource.java was approved and merged into the cbioportal codebase previously with https://github.com/cBioPortal/cbioportal/pull/10424

This PR also cleans up the code, providing functions for:
- logging warnings for use of deprecated datasource properties
- validating the presence of 4 required datasource properties (spring.datasource.url was added)
- stringIsNullOrBlank

According to the final section notes of https://docs.oracle.com/javase/specs/jls/se7/html/jls-14.html#jls-14.10 it is bad practice to handle invalid argument problems through assert() because assert processing is disabled by default but validity of arguments should be checked regardless of whether assertions are enabled or not. Also, an appropriate exception is not thrown by assert(), which causes an AssertionError to be thrown instead. For these reasons, assert() was replaced by throw new RuntimeException(), which causes the construction of the datasource bean to fail during the bean construction stage of context creation, exiting the application.

Note : this PR replaced #26 because currently, pull requests across forks seems to cause a failure in the CI tests (updating dependencies in the manifest / maven.yml)

### DIFF
--- a/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
@@ -1,48 +1,71 @@
 package org.mskcc.cbio.portal.dao;
 
+import java.util.ArrayList;
+import java.util.List;
 import org.apache.commons.dbcp2.BasicDataSource;
 import org.mskcc.cbio.portal.util.DatabaseProperties;
-import org.apache.commons.lang3.StringUtils;
-import org.springframework.util.Assert;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


mandawilson (approved):
Thanks Rob!

PR #28, opened 2024-03-18
update mysql-connector-java to 8.0.28

sheridancbio:
The cbioportal codebase currently uses mysql-connector-java version 8.0.28, and cbioportal is brought into cbioportal-core as a dependency. This PR updates the mysql-connector-java version in cbioportal-core to match the one cbioportal.

Two connected changes were needed:
- the updated mysql-connector-java is more strict, disallowing any call of ResultSet.first() if the (default) ResultSet type is FORWARD_ONLY. Such a call was changed to ResultSet.next() to move the cursor onto the first record.
- the updated mysql-connector-java is more reluctant to allow bulk upload of data from local files (LOAD DATA LOCAL INFILE ...), so an additional property (allowLoadLocalInfile) needed to be set in the defined DataSource when opening connections to the database.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
 		<maf.version>1.0.0</maf.version>
 
 
-		<mysql_driver.version>5.1.48</mysql_driver.version>
+		<mysql_driver.version>8.0.28</mysql_driver.version>
 		<mysql_commons_lang3.version>3.13.0</mysql_commons_lang3.version>
 
 		<commons_dbcp2.version>2.10.0</commons_dbcp2.version>
--- a/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/JdbcDataSource.java
@@ -38,7 +38,7 @@ public class JdbcDataSource extends BasicDataSource {
         }
     }
 
-    public JdbcDataSource () {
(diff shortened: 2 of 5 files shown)


sheridancbio:
This PR is still being tested.

sheridancbio:
This PR has been tested in a built version of the msk importer. A typical study was imported with a variety of genetic profiles, and importation was successful. Similar changes to the datasource (allowLoadLocalInfile) were made in the msk importer codebase.

sheridancbio:
@JREastonMarks I think this could be merged, but I also have not tested an import with the metaImport.py script (the msk importer performs the import using direct calls to methods of the java classes in the scripts package). So if anyone would like to perform such a test, or tell me an easy way to use metaImport.py to import a representative test study (I've not done it before) I'd be happy to run that test.

JREastonMarks (approved):

sheridancbio:
This PR was merged after successful test of an adapted version of the test_load_study.sh script from https://github.com/cBioPortal/cbioportal/blob/master/test/integration/test_load_study.sh
which did not require docker download of a released version of cbioportal-core. Instead it relied on a standalone database and a standalone cbioportal deployment on a specified port. The cbioportal deployment used the codebase at 8361b0a3dca63050f134a9ab7caa5002bdaab63a (approximately for frontend release v6.0.6) and the db_schema_version was 2.13.1

PR #38, opened 2024-05-15
Implement incremental upload of tab delimited data. 

forus:
See https://github.com/cBioPortal/cbioportal-core/pull/45 instead.


src/main/java/org/mskcc/cbio/portal/dao/DaoCnaEvent.java
```diff
@@ -119,7 +119,32 @@ private static long addCnaEventDirectly(CnaEvent cnaEvent) throws DaoException {
             JdbcUtil.closeAll(DaoCnaEvent.class, con, pstmt, rs);
         }
 	}
-    
+
+    public static void removeSampleCnaEvents(int cnaProfileId, List<Integer> sampleIds) throws DaoException {
+        Connection con = null;
+        PreparedStatement pstmt = null;
+        ResultSet rs = null;
+        try {
```
haynescd, line 127: Is there a reason you don't use try with resources?

haynescd (commented):
just had a couple questions, but looks good

PR #42, opened 2024-06-04
(4/7) RFC79: Implement incremental upload of CNA DISCRETE long data

forus:
Please review https://github.com/cBioPortal/cbioportal-core/pull/41 next.

_Please note, to avoid the costly process of rebasing all pull requests (PRs), all feedback will be incorporated into the final PR._


src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
```diff
@@ -0,0 +1,111 @@
...
+    public boolean store(int geneticEntityId, String[] values) throws DaoException {
+        daoGeneticAlteration.deleteAllRecordsInGeneticProfile(geneticProfileId, geneticEntityId);
+        String[] expandedValues = extendValues(geneticEntityId, values);
+        return geneticAlterationImporter.store(geneticEntityId, expandedValues);
+    }
+
+    @Override
+    public void finalise() {
+        expandRemainingGeneticEntityTabDelimitedRowsWithBlankValue();
+        geneticAlterationImporter.finalise();
```
pieterlukasse, line 70: why this last line?

src/main/java/org/mskcc/cbio/portal/scripts/ImportCnaDiscreteLongData.java
```diff
@@ -22,52 +22,108 @@
...
+        if (!Set.of(DISCRETE.name(), DISCRETE_LONG.name()).contains(geneticProfile.getDatatype())) {
+            throw new IllegalStateException("Platform "
+                    + geneticProfileId
+                    + " has not supported datatype: "
+                    + geneticProfile.getDatatype());
+        }
         this.genePanel = genePanel;
         this.daoGene = daoGene;
-        this.geneticAlterationGeneImporter = new GeneticAlterationImporter(geneticProfileId, daoGeneticAlteration);
+        this.updateMode = updateMode;
```
pieterlukasse, line 102:
```suggestion
        this.isIncrementalUpdateMode = isIncrementalUpdateMode;
```
I keep tripping over this 😅 ...  just `updateMode` does not sound like a boolean to me, but rather like it might contain the mode in which the current data update is running...

src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
```diff
@@ -83,10 +86,18 @@ public static GeneticProfile loadGeneticProfile(File file) throws IOException, D
...
+                    throw new IllegalStateException("genetic_profile record found with same Stable ID ("
+                            + existingGeneticProfile.getStableId() + ") but unsupported data type: "
+                            + existingGeneticProfile.getDatatype());
+                }
+            } else {
+                if (!existingGeneticProfile.getDatatype().equals(geneticProfile.getDatatype())) {
+                    throw new IllegalStateException("genetic_profile record found with same Stable ID ("
+                            + existingGeneticProfile.getStableId() + ") but different data type: "
+                            + existingGeneticProfile.getDatatype());
+                }
```
pieterlukasse, line 100: what is supported or not is checked elsewhere, no? Maybe here it would suffice to just have the last check?

pieterlukasse (commented):
thanks @forus 
some comments / questions added

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
forus: `geneticAlterationImporter.finalise()` does nothing atm. So removing the line won't change the behavior now. I wanted to call the finalise to notify the wrapped class to clean things, in case such logic appears there.

Let me know if you think we should remove the call.

src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
forus: I think it's unfortunate how `DISCRETE_LONG` has been implemented.

`DISCRETE_LONG` is not a separate data type, but an input file shape. We have to remember that `DISCRETE_LONG` and `DISCRETE` are the same data types at the end while doing this validation.

src/main/java/org/mskcc/cbio/portal/scripts/ImportCnaDiscreteLongData.java
forus: see here https://github.com/cBioPortal/cbioportal-core/pull/47/commits/79c404129862cf39800091a6046960e9a680f101

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
pieterlukasse: should we use `super.finalise()` instead? I think that would make more sense. Same for the other methods in this class like `store`

forus: @pieterlukasse please note `GeneticAlterationIncrementalImporter` is not a superclass of `GeneticAlterationImporterImpl`. I intentionally use composition over inheritance here to address the fragile base class problem.

pieterlukasse: got it. I still think writing it in the conventional way through inheritance makes the code easier to follow.

forus: done https://github.com/cBioPortal/cbioportal-core/commit/d98392a291a3ace8afb137c3649ed3c272166fb5

pieterlukasse: should `super.` be replaced by `this.` in a number of assignments?

forus: It can. It's a good question whether it should. I thought using super in the constructor communicates better that's an inherited field.

forus: here is the change https://github.com/cBioPortal/cbioportal-core/pull/47/commits/96acec541d3afec7ac19f189cb1cd0a471dbcafc

haynescd (approved):
LGTM

PR #44, opened 2024-06-04
(2/7) RFC79: Make study_es_0_inc data pass validation

forus:
- document how to load `study_es_0` study
- Remove `SpringUtil.initDataSource();` from multiple classes as it was causing not used services to fail during initialisation
-  organise imports (remove unused, no wildcard imports) for the classes that lost `SpringUtil.initDataSource();` statement.
- include into the test data (`seed_mini.sql`) genes that are used by `study_es_0` data.

Please review https://github.com/cBioPortal/cbioportal-core/pull/43 next.

Please note, to avoid the costly process of rebasing all pull requests (PRs), all feedback will be incorporated into the final PR.

PR #43, opened 2024-06-04
(3/7) RFC79: Implement incremental upload for timeline data

forus:
Please review https://github.com/cBioPortal/cbioportal-core/pull/42 next.

Please note, to avoid the costly process of rebasing all pull requests (PRs), all feedback will be incorporated into the final PR.

PR #40, opened 2024-06-04
(6/7) RFC79: Implement incremental upload of structural variants data

forus:
Please review https://github.com/cBioPortal/cbioportal-core/pull/39 next.

_Please note, to avoid the costly process of rebasing all pull requests (PRs), all feedback will be incorporated into the final PR._

PR #39, opened 2024-06-04
(7/7) RFC79: Implement incremental upload of CNA segmented data

forus:
Reminder: After merging this PR, incorporate all feedback in a new PR.
After it's merged, create `rfc79` -> `main` PR and remove "do not merge" from this one https://github.com/cBioPortal/cbioportal/pull/10816

_Please note, to avoid the costly process of rebasing all pull requests (PRs), all feedback will be incorporated into the final PR._

PR #48, opened 2024-06-19
RFC79: Incremental Upload of Data Entries

forus:
See also [RFC79](https://docs.google.com/document/d/1oTKtC4GtDNRrHscMIKGRrx82o1sUoinjWvspnMJIOv4/edit)

The solution involves extending the current `metaImport.py` script and java data loader commands with additional flags to support the incremental upload of entries. This approach allows users to add patients, samples, and molecular data without having to reupload the entire study.

Read more in docs https://github.com/cBioPortal/cbioportal/pull/10816

To review in logical parts and see previous discussions about the implementation, you might want to check the chain of closed PRs this PR consists of:

- [Incremental data upload for mutations, case lists, patient and sample attributes](https://github.com/cBioPortal/cbioportal-core/pull/32)
- [(1/7) RFC 79: Implement incremental upload of tab delimited data.](https://github.com/cBioPortal/cbioportal-core/pull/45)
- [(2/7) RFC79: Make study_es_0_inc data pass validation](https://github.com/cBioPortal/cbioportal-core/pull/44)
- [(3/7) RFC79: Implement incremental upload for timeline data](https://github.com/cBioPortal/cbioportal-core/pull/43)
- [(4/7) RFC79: Implement incremental upload of CNA DISCRETE long data](https://github.com/cBioPortal/cbioportal-core/pull/42)
- [(5/7) RFC79: Implement incremental upload for gene panel matrix](https://github.com/cBioPortal/cbioportal-core/pull/41)
- [(6/7) RFC79: Implement incremental upload of structural variants data](https://github.com/cBioPortal/cbioportal-core/pull/40)
- [(7/7) RFC79: Implement incremental upload of CNA segmented data](https://github.com/cBioPortal/cbioportal-core/pull/39)
- [RFC79: Implement feedback from previous 7 PRs](https://github.com/cBioPortal/cbioportal-core/pull/47)


src/main/java/org/mskcc/cbio/portal/dao/DaoSampleList.java
```diff
@@ -54,15 +56,22 @@ public int addSampleList(SampleList sampleList) throws DaoException {
...
             pstmt.setString(4, sampleList.getSampleListCategory().getCategory());
             pstmt.setString(5, sampleList.getDescription());
             rows = pstmt.executeUpdate();
-   			int listListRow = addSampleListList(sampleList, con);
-   			rows = (listListRow != -1) ? (rows + listListRow) : rows;
+            try (ResultSet generatedKey = pstmt.getGeneratedKeys()) {
+                if (generatedKey.next()) {
+                    int listId = generatedKey.getInt(1);
+                    int listListRow = addSampleListList(sampleList.getCancerStudyId(), listId, sampleList.getSampleList(), con);
+                    rows = (listListRow != -1) ? (rows + listListRow) : rows;
```
pieterlukasse, line 70: when will `addSampleListList` return -1 ? It looks like this scenario was removed from there.

src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
```diff
@@ -68,27 +95,37 @@ public class ImportExtendedMutationData{
...
     public ImportExtendedMutationData(File mutationFile, int geneticProfileId, String genePanel, Set<String> filteredMutations, Set<String> namespaces) {
+        this(mutationFile, geneticProfileId, genePanel, filteredMutations, namespaces, false);
+    }
+
+    public ImportExtendedMutationData(File mutationFile, int geneticProfileId, String genePanel, Set<String> filteredMutations, Set<String> namespaces, boolean overwriteExisting) {
         this.mutationFile = mutationFile;
         this.geneticProfileId = geneticProfileId;
         this.swissprotIsAccession = false;
-        this.genePanel = genePanel;
+        this.genePanelId = (genePanel == null) ? null : GeneticProfileUtil.getGenePanelId(genePanel);;
```
pieterlukasse, line 122: double ;;

src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
```diff
@@ -563,53 +602,35 @@ private String transformOMAScore( String omaScore) {
...
         while (line.startsWith("#")) {
             Matcher seqSamplesMatcher = SEQUENCE_SAMPLES_REGEX.matcher(line);
             // line is of format #sequenced_samples: STABLE_ID STABLE_ID STABLE_ID STABLE_ID
             if (seqSamplesMatcher.find()) {
-                addSampleProfileRecords(getSequencedSamples(seqSamplesMatcher.group(1), geneticProfile));
+                internalSampleIds.addAll(getSequencedInternalSampleId(seqSamplesMatcher.group(1), geneticProfile));
             }
             line = buffer.readLine().trim();
         }
+        DaoSampleProfile.upsertSampleProfiles(internalSampleIds, geneticProfileId, genePanelId);
```
pieterlukasse, line 614: I think upsertSampleProfileRecords is a better name. Or upsertSampleToProfileRecords , or upsertSampleVsProfileRecords . "Sampleprofiles" is confusing as it sounds like it is referring to the complete sample profile (i.e. the mutation data itself).

src/main/java/org/mskcc/cbio/portal/scripts/ImportProfileData.java
```diff
@@ -131,22 +131,23 @@ public void run() {
...
-                        daoGeneticAlteration
+                        overwriteExisting,
+                        daoGene
                     );
-                    genericAssayProfileImporter.importData(numLines);
+                    genericAssayProfileImporter.importData();
                 }
             } else if(
                 geneticProfile.getGeneticAlterationType() == GeneticAlterationType.COPY_NUMBER_ALTERATION 
-                && DISCRETE_LONG.name().equals(geneticProfile.getDatatype())
```
pieterlukasse, line 140: why `geneticProfile.getDatatype()` -> `geneticProfile.getOtherMetaDataField("datatype")` ?

src/main/java/org/mskcc/cbio/portal/scripts/ImportSif.java
pieterlukasse, line 1: this class is not doing anything. Let's remove it completely.

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
```diff
@@ -0,0 +1,105 @@
...
+            }
+            if (sampleIdToValue.containsKey(sampleId)) {
+                updatedSampleValues[i] = sampleIdToValue.get(sampleId);
+            }
+        }
+        return updatedSampleValues;
+    }
+
+    private Map<Integer, String> mapWithFileOrderedSampleList(String[] values) {
+        return ArrayUtil.zip(fileOrderedSampleList.toArray(new Integer[0]), values);
```
pieterlukasse, line 90: maybe Integer[]::new

src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
pieterlukasse: upsertSampleToProfileMapping is perhaps the best name...

src/main/java/org/mskcc/cbio/portal/util/CaseListReader.java
```diff
@@ -0,0 +1,43 @@
...
+            throw new RuntimeException(e);
+        }
+
+        String stableId = properties.getProperty("stable_id");
+        String cancerStudyIdentifier = properties.getProperty("cancer_study_identifier");
+        String caseListName = properties.getProperty("case_list_name");
+        String caseListDescription = properties.getProperty("case_list_description");
+        String caseListCategory = properties.getProperty("case_list_category");
+        String caseListIds = properties.getProperty("case_list_ids");
+        List<String> sampleIds = caseListIds == null ? List.of()
```
pieterlukasse, line 29: I think case_list_ids is mandatory, so we should throw error if null

src/main/java/org/mskcc/cbio/portal/util/TsvUtil.java
```diff
@@ -0,0 +1,43 @@
+package org.mskcc.cbio.portal.util;
+
+/**
+ * Utils to parse and validate TSV lines
+ * @author Ruslan Forostianov
+ */
+public class TsvUtil {
+    /**
+     * is the line has some data
+     * e.g. blank line and comments do not
```
pieterlukasse, line 10: comments seem incomplete

pieterlukasse (commented):
thanks @forus . Please take a look at my final comments. 

Review note: I've skipped over the test files.

src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
forus: Removed https://github.com/cBioPortal/cbioportal-core/pull/48/commits/df8f7aff66ddba603a7ae3276eceec7c4f7015d6#diff-b3cccdbc78574935ab64e4df548ce86deb18cf7d2ca0967597862cc56a533c22R122

src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
forus: I like `upsertSampleToProfileMapping` Renamed here https://github.com/cBioPortal/cbioportal-core/pull/48/commits/f120f5d14e4c2928309d89a041c8fdfb1920476a

src/main/java/org/mskcc/cbio/portal/scripts/ImportProfileData.java
forus: `DISCRETE_LONG` ends up in the database as `DISCRETE` at the end.

In the case when the genetic profile has been loaded to the database previously, we reuse the platform and we want to access how the datatype has been defined in the file with `geneticProfile.getOtherMetaDataField("datatype")` instead of how it is stored in the database `geneticProfile.getDatatype()`

src/main/java/org/mskcc/cbio/portal/scripts/ImportSif.java
forus: here is separate issue for that https://github.com/cBioPortal/cbioportal-core/issues/49

src/main/java/org/mskcc/cbio/portal/util/CaseListReader.java
forus: It's done in [CaseListValidator](https://github.com/cBioPortal/cbioportal-core/blob/rfc79/src/main/java/org/mskcc/cbio/portal/validate/CaseListValidator.java#L24). The reader is not concerned with what fields are compulsory in the current design.

src/main/java/org/mskcc/cbio/portal/util/TsvUtil.java
forus: Made complete https://github.com/cBioPortal/cbioportal-core/pull/48/commits/28dfa054af695c215a97b5bfbaedf53e92f4d9bb

src/main/java/org/mskcc/cbio/portal/scripts/UpdateCaseListsSampleIds.java
```diff
@@ -0,0 +1,213 @@
...
+    private LinkedHashSet<String> allSampleIds;
+
+    public UpdateCaseListsSampleIds(String[] args) {
+        super(args);
+    }
+
+    /**
+     * Updates case list sample ids from clinical sample and case list files
+     */
+    public void run() {
```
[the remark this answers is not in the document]
forus: Here are comments for `updateCaseLists` https://github.com/cBioPortal/cbioportal-core/pull/48/commits/28dfa054af695c215a97b5bfbaedf53e92f4d9bb

Please let me know which other pieces of code require more explanation in comments

forus:
@pieterlukasse Thanks! Please have another look.

pieterlukasse (dismissed):

sheridancbio:
I will try to review this.

One quick comment - when it comes time to merge this PR I hope we use "squash and merge" rather than merging in the individual changesets.

.github/workflows/validate-python.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: 'Validate tests'
         working-directory: ./cbioportal-core
         run: |
-          docker run -v ${PWD}:/cbioportal-core python:3.6 /bin/bash -c '
+          docker run -v ${PWD}:/cbioportal-core python:3.6 /bin/sh -c '
             cd cbioportal-core &&
             pip install -r requirements.txt &&
-            source test_scripts.sh'
+            ./test_scripts.sh'
```
sheridancbio, line 20: This may not matter, but running test_scripts.sh like this invokes it in a subshell rather than in the current (shell) process.
If there were environment variables set by the script which need to be exported or read out they would be lost this way.

forus: indeed. If we lose any relevant env. variable, tests will fail.
`source` command is bash specific btw

scripts/importer/validateData.py
```diff
@@ -727,7 +727,7 @@ def checkSampleId(self, sample_id, column_number):
 
         Return True if the sample id was valid, False otherwise.
         """
-        if sample_id not in DEFINED_SAMPLE_IDS:
+        if DEFINED_SAMPLE_IDS is not None and sample_id not in DEFINED_SAMPLE_IDS:
```
sheridancbio, line 730: I guess these changes were needed because now an incremental import might not contain any samples. Or might not contain any patients, etc.
But in the "import whole study" mode, should we try to detect (and reject) a study which had no patients (and therefore no samples)? Or should we allow it to go in?
Maybe this is covered in the way that DEFINED_SAMPLE_IDS is initialized somewhere (None for incremental, but some empty container for whole study import)?

scripts/importer/validateData.py
```diff
@@ -5341,7 +5340,6 @@ def validate_study(study_dir, portal_instance, logger, relaxed_mode, strict_maf_
     attributes are not None.
     """
 
-    global DEFINED_CANCER_TYPES
```
sheridancbio, line 5344: dead code?

scripts/importer/validateData.py
```diff
@@ -5644,7 +5667,10 @@ def main_validate(args):
     # set portal version
     cbio_version = portal_instance.portal_version
 
-    validate_study(study_dir, portal_instance, logger, relaxed_mode, strict_maf_checks)
+    if partial_data:
+        validate_data_dir(data_dir, portal_instance, logger, relaxed_mode, strict_maf_checks)
+    else:
+        validate_study(data_dir, portal_instance, logger, relaxed_mode, strict_maf_checks)
```
sheridancbio, line 5673: Maybe this addresses my concern above about validating "whole study import" ... there is a separate method here, so maybe the validation approach for whole study import is different.

sheridancbio (commented):
Partial review (python components only)
Looking good so far.

src/main/java/org/mskcc/cbio/portal/dao/DaoSampleList.java
```diff
@@ -248,30 +226,46 @@ private int addSampleListList(SampleList sampleList, Connection con) throws DaoE
...
+    public void updateSampleListList(SampleList sampleList) throws DaoException {
+        Connection con = null;
+        PreparedStatement pstmt = null;
+        try {
+            con = JdbcUtil.getDbConnection(DaoSampleList.class);
+            pstmt = con.prepareStatement(DELETE_SAMPLE_LIST_LIST);
+            pstmt.setInt(1, sampleList.getSampleListId());
+            pstmt.executeUpdate();
+
+            addSampleListList(sampleList.getCancerStudyId(), sampleList.getSampleListId(), sampleList.getSampleList(), con);
```
sheridancbio, line 260: This is a complete delete / replace for the records listing the samples in a sample list.

src/main/java/org/mskcc/cbio/portal/dao/DaoSampleProfile.java
```diff
@@ -50,98 +60,62 @@ public final class DaoSampleProfile {
...
-            } else {
-                MySQLbulkLoader.getMySQLbulkLoader(TABLE_NAME).insertRecord(
-                    Integer.toString(sampleId),
-                    Integer.toString(geneticProfileId),
-                    null);
-            }
 
-            return 1;
-        }
+    public static void upsertSampleToProfileMapping(Collection<Integer> sampleIds, Integer geneticProfileId, Integer panelId) throws DaoException {
```
sheridancbio, line 64: There are not too many records involved here (# of samples) * (# of profiles), but I do wonder whether switching from the bulk loader to individual SQL statements will have a performance impact. I guess we will see.

src/main/java/org/mskcc/cbio/portal/dao/DaoSampleProfile.java
```diff
@@ -50,98 +60,62 @@ public final class DaoSampleProfile {
...
-                }
-                else {
-                    pstmt.setNull(3, java.sql.Types.INTEGER);
+            con = JdbcUtil.getDbConnection(DaoSampleProfile.class);
+
+            pstmt = con.prepareStatement
+                    ("INSERT INTO sample_profile (`SAMPLE_ID`, `GENETIC_PROFILE_ID`, `PANEL_ID`)" +
+                            " VALUES" +
+                            String.join(",", Collections.nCopies(idTuples.size(), " (?,?,?)")) +
+                            " ON DUPLICATE KEY UPDATE `PANEL_ID` = VALUES(`PANEL_ID`);");
```
sheridancbio, line 85: I see ... this is overwriting the panel_id for the existing records.

src/main/java/org/mskcc/cbio/portal/dao/DaoSampleProfile.java
```diff
@@ -160,14 +134,20 @@ public static boolean sampleExistsInGeneticProfile(int sampleId, int geneticProf
         try {
             con = JdbcUtil.getDbConnection(DaoSampleProfile.class);
             pstmt = con.prepareStatement
-                    ("SELECT * FROM sample_profile WHERE SAMPLE_ID = ? AND GENETIC_PROFILE_ID = ?");
+                    ("SELECT PANEL_ID FROM sample_profile WHERE SAMPLE_ID = ? AND GENETIC_PROFILE_ID = ?");
             pstmt.setInt(1, sampleId);
             pstmt.setInt(2, geneticProfileId);
             rs = pstmt.executeQuery();
-            return (rs.next());
```
sheridancbio, line 167: This seems like the function was broken before. Is this a bugfix? Or was this function previously only providing functionality of "sampleXInProfileYDoesHaveAPanelIdIdentified" (boolean)

sheridancbio (commented):
partial review (covering the DAO classes only)

pom.xml
```diff
@@ -252,6 +252,9 @@
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-surefire-plugin</artifactId>
 				<version>2.21.0</version>
+				<configuration>
+					<trimStackTrace>false</trimStackTrace>
```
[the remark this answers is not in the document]
forus: yes, for debugging. It'd be handy to see where the issue occurred in the custom code when tests fail.

scripts/importer/cbioportalImporter.py
```diff
@@ -212,11 +228,20 @@ def process_command(jvm_args, command, meta_filename, data_filename, study_ids,
...
+        import_data(jvm_args, meta_filename, data_filename, update_generic_assay_entity)
     elif command == IMPORT_CASE_LIST:
         import_case_list(jvm_args, meta_filename)
 
-def process_directory(jvm_args, study_directory, update_generic_assay_entity = None):
+def get_meta_filenames(data_directory):
+    meta_filenames = [
+        os.path.join(data_directory, meta_filename) for
+        meta_filename in os.listdir(data_directory) if
+        re.search(r'(\b|_)meta(\b|[_0-9])', meta_filename,
```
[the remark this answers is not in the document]
forus: It has to fail later to parse the data file as a metafile. Not sure how friendly that error message would be. We could improve that error message by adding a reminder that having _meta_ in the name of a data file is not a good idea.
This issue hasn't been introduced by the PR. I've just extracted this piece of code into a separate function.

Anyway, please feel free to report an issue though. I think it's a valid improvement.

scripts/importer/validateData.py
forus: In "import whole study" mode:
   - if `--relaxed_clinical_definitions` is set, `DEFINED_SAMPLE_IDS` would be equal to empty set (`set()`, not `None`) and fail the validation in this line
   - otherwise, it'd fail even earlier here https://github.com/cBioPortal/cbioportal-core/blob/rfc79/scripts/importer/validateData.py#L5430

scripts/importer/validateData.py
forus: this variable is used locally only, no need to keep it global. You helped me to spot, that  I forgot to remove it from here https://github.com/cBioPortal/cbioportal-core/blob/rfc79/scripts/importer/validateData.py#L73

forus: I forgot to remove it from more places. Please review this commit https://github.com/cBioPortal/cbioportal-core/pull/48/commits/a8f014aaeb91a71bf87bd5b961e2ee4317d16867

src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -61,7 +73,6 @@ public static enum Status {
     private static final Map<Integer,CancerStudy> byInternalId = new HashMap<Integer,CancerStudy>();
 
     static {
-        SpringUtil.initDataSource();
```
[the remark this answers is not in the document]
forus: I guess you set transactional data source like here https://github.com/cBioPortal/cbioportal-core/blob/rfc79/src/test/resources/applicationContext-dao.xml#L26 If yes, removing the initialisation of spring context here should not break anything.

but indeed, please test the MSK importer after taking this change to be sure it'd not break anything.

src/main/java/org/mskcc/cbio/portal/dao/DaoClinicalData.java
```diff
@@ -361,7 +364,48 @@ public static List<ClinicalData> getSampleData(int cancerStudyId, Collection<Str
...
+        Connection con = null;
+        PreparedStatement pstmt = null;
+        try {
+            con = JdbcUtil.getDbConnection(DaoClinicalData.class);
+            pstmt = con.prepareStatement("DELETE FROM " + SAMPLE_ATTRIBUTES_TABLE
+                    + " WHERE `ATTR_ID` = ? AND `INTERNAL_ID` IN ("
+                    + String.join(",", Collections.nCopies(sampleInternalIds.size(), "?"))
+                    + ")");
+            int parameterIndex = 1;
+            pstmt.setString(parameterIndex++, attrId);
```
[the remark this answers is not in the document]
forus: haha, indeed) let me know if you want me to change this.

src/main/java/org/mskcc/cbio/portal/dao/DaoSampleProfile.java
forus: The function worked fine. It's not a bug fix. It's just removing the code that is not used anymore. Switching to `upsertSampleToProfileMapping` made this check not needed.

src/main/java/org/mskcc/cbio/portal/dao/DaoClinicalData.java
```diff
@@ -361,7 +364,48 @@ public static List<ClinicalData> getSampleData(int cancerStudyId, Collection<Str
...
+            for (Integer sampleInternalId : sampleInternalIds) {
+                pstmt.setInt(parameterIndex++, sampleInternalId);
+            }
+            pstmt.executeUpdate();
```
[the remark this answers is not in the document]
forus: Please see here https://github.com/cBioPortal/cbioportal-core/pull/48/commits/0df91c923d996fd1e4d1c4a0c1c93b929f46abfa

src/main/java/org/mskcc/cbio/portal/dao/DaoCopyNumberSegment.java
```diff
@@ -80,8 +80,15 @@ public static void createFractionGenomeAlteredClinicalData(int cancerStudyId) th
                     "AS c2 WHERE c2.`CANCER_STUDY_ID` = c1.`CANCER_STUDY_ID` AND c2.`SAMPLE_ID` = c1.`SAMPLE_ID` AND " + 
                     "ABS(c2.`SEGMENT_MEAN`) >= 0.2) / SUM(`END`-`START`)) AS `VALUE` FROM `copy_number_seg` AS c1 , `cancer_study` " +
                     "WHERE c1.`CANCER_STUDY_ID` = cancer_study.`CANCER_STUDY_ID` AND cancer_study.`CANCER_STUDY_ID`=? " +
-                    "GROUP BY cancer_study.`CANCER_STUDY_ID` , `SAMPLE_ID` HAVING SUM(`END`-`START`) > 0;");
-            pstmt.setInt(1, cancerStudyId);
+                            (sampleIds == null ? "" : ("AND `SAMPLE_ID` IN ("+ String.join(",", Collections.nCopies(sampleIds.size(), "?")) + ") "))
```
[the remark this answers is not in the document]
forus: Please review here https://github.com/cBioPortal/cbioportal-core/pull/48/commits/990218d41f943ccb60b3105495607cf128087b21

src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -475,18 +477,10 @@ public static ArrayList<ExtendedMutation> getMutations (long entrezGeneId, Strin
         return mutationList;
     }
 
-    /**
```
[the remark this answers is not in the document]
forus: yes, both functions are used within the codebase. I found them handy for new logic and tests.

src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -1501,18 +1495,20 @@ protected static String boolToStr(boolean value)
         return value ? "1" : "0";
     }
 
-    /**
-     * @deprecated  We believe that this method is no longer called by any part of the codebase, and it will soon be deleted.
-     */
-    @Deprecated
-    public static void deleteAllRecordsInGeneticProfile(long geneticProfileId) throws DaoException {
+    public static void deleteAllRecordsInGeneticProfileForSample(long geneticProfileId, long internalSampleId) throws DaoException {
```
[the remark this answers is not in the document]
forus: I haven't found any code using `public static void deleteAllRecordsInGeneticProfile(long geneticProfileId) throws DaoException`.

Removing mutations when the study is being removed happens due to the chain of cascade deleting foreign keys, right?

src/main/java/org/mskcc/cbio/portal/dao/DaoSampleList.java
```diff
@@ -42,31 +42,38 @@
  */
 public class DaoSampleList {
 
-	/**
+    public static final String DELETE_SAMPLE_LIST_LIST = "DELETE FROM sample_list_list WHERE `LIST_ID` = ?";
```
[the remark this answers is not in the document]
forus: Agree https://github.com/cBioPortal/cbioportal-core/pull/48/commits/2bbe3c5c261b1e74766394a3ba46424250a7d2af

src/main/java/org/mskcc/cbio/portal/dao/DaoSampleList.java
```diff
@@ -42,31 +42,38 @@
...
             pstmt.setString(3, sampleList.getName());
             pstmt.setString(4, sampleList.getSampleListCategory().getCategory());
             pstmt.setString(5, sampleList.getDescription());
             rows = pstmt.executeUpdate();
-   			int listListRow = addSampleListList(sampleList, con);
-   			rows = (listListRow != -1) ? (rows + listListRow) : rows;
+            try (ResultSet generatedKey = pstmt.getGeneratedKeys()) {
+                if (generatedKey.next()) {
+                    int listId = generatedKey.getInt(1);
+                    rows += addSampleListList(sampleList.getCancerStudyId(), listId, sampleList.getSampleList(), con);
```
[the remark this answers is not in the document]
forus: I agree. This match looks weird. Let's not do this https://github.com/cBioPortal/cbioportal-core/pull/48/commits/bc76d3b03235298e6bc6c63d172bc2138c8bd9e9

src/main/java/org/mskcc/cbio/portal/dao/DaoSampleList.java
forus: Do you anticipate any problems by doing an update with delete+insert?

src/main/java/org/mskcc/cbio/portal/dao/DaoCnaEvent.java
```diff
@@ -119,7 +119,34 @@ private static long addCnaEventDirectly(CnaEvent cnaEvent) throws DaoException {
...
+                            " FROM sample_cna_event" +
+                            " LEFT JOIN alteration_driver_annotation ON alteration_driver_annotation.`ALTERATION_EVENT_ID` = sample_cna_event.`CNA_EVENT_ID`" +
+                            " AND alteration_driver_annotation.`SAMPLE_ID` = sample_cna_event.`SAMPLE_ID`" +
+                            " AND alteration_driver_annotation.`GENETIC_PROFILE_ID` = sample_cna_event.`GENETIC_PROFILE_ID`" +
+                            " WHERE sample_cna_event.`GENETIC_PROFILE_ID` = ? AND sample_cna_event.`SAMPLE_ID` IN (" +
+                            String.join(",", Collections.nCopies(sampleIds.size(), "?"))
+                            + ")");
+            pstmt.setInt(1, cnaProfileId);
+            for (int i = 0; i < sampleIds.size(); i++) {
+                pstmt.setInt(i + 2, sampleIds.get(i));
```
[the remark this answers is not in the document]
forus: I chose "c style". The change is here https://github.com/cBioPortal/cbioportal-core/pull/48/commits/0c0b754d1f5e526a980d6522f1c81ee02b456b0a

src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticAlteration.java
```diff
@@ -238,9 +236,21 @@ public HashMap<Integer,HashMap<Integer, String>> getGeneticAlterationMapForEntit
...
                 String values = rs.getString("VALUES");
-                //hm.debug..
-                String valueParts[] = values.split(DELIM);
-                for (int i=0; i<valueParts.length; i++) {
+                String[] valueParts = values.split(DELIM, -1);
+                int valuesLength = valueParts.length;
+                boolean hasMeaninglessTrailingDelimiter = valuesLength - orderedSampleList.size() == 1 && valueParts[valuesLength - 1].isEmpty();
+                if (hasMeaninglessTrailingDelimiter) {
+                    // adjust value length to account for the trailing delimiter
+                    valuesLength -= 1;
```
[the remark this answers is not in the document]
forus: We compensate for this code https://github.com/cBioPortal/cbioportal-core/blob/rfc79/src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticAlteration.java#L107
that forms csv row with a trailing comma. Even if we fix that code, DBs out there would contain rows with the trailing comma still. So this code is written to work with rows with trailing commas and without.

BTW I've just realized that I don't know about the piece of code that would produce rows without trailing commas. So maybe we should simplify this logic and always expect CSV rows to have a trailing comma. Let me know what you think.

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationImporter.java
```diff
@@ -2,25 +2,40 @@
...
-        int geneticProfileId, 
-        DaoGeneticAlteration daoGeneticAlteration
-    ) {
+        int geneticProfileId,
+        List<Integer> orderedSampleList
+    ) throws DaoException {
         this.geneticProfileId = geneticProfileId;
-        this.daoGeneticAlteration = daoGeneticAlteration;
+        this.orderedSampleList = orderedSampleList;
+        storeOrderedSampleList();
```
sheridancbio, line 31: This makes construction of this class dependent on a successful database operation. In general I think this is not recommended. Could this function call be moved inside of the store() methods below? Also, if the storage of the sample id list succeeds here, but the storage of the cna event profile fails during store() would we have a problem? (or maybe transactions will take care of that?)

I suppose that each instance of this class is specific to a particular genetic alteration profile, so we do not need to worry about reuse of any instance to import a different profile (and maybe need to store a different set of sample ids)

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationImporter.java
```diff
@@ -29,31 +44,59 @@ public GeneticAlterationImporter(
...
+
+
+    /**
+     * Universal method that stores values for different genetic entities
+     * @param geneticEntityId
+     * @param values
+     * @return true if entity has been stored, false - if entity already existed
+     * @throws DaoException
+     */
+    public boolean store(
```
sheridancbio, line 77: This is interesting. Either we did not previously support the import of genetic entity profiles, or this functionality was provided elsewhere (perhaps it is moving here now)

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
```diff
@@ -0,0 +1,105 @@
...
+        this.geneticAlterationMap = daoGeneticAlteration.getGeneticAlterationMapForEntityIds(geneticProfileId, null);
+        this.fileOrderedSampleList = fileOrderedSampleList;
+
+        ArrayList <Integer> savedOrderedSampleList = DaoGeneticProfileSamples.getOrderedSampleList(this.geneticProfileId);
+        int initialOrderSampleListSize = savedOrderedSampleList.size();
+        geneticAlterationMap.forEach((geneticEntityId, sampleToValue) -> {
+            if (sampleToValue.size() != initialOrderSampleListSize) {
+                throw new IllegalStateException("Number of samples ("
+                        + sampleToValue.size() + ") for genetic entity with id "
+                        + geneticEntityId + " does not match with the number in the inital sample list ("
```
sheridancbio, line 37: I think "initial" could be misinterpreted. Perhaps "preexisting" or "established" would be more accurate?

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
```diff
@@ -0,0 +1,105 @@
...
+                        + initialOrderSampleListSize + ").");
+            }
+        });
+        // add all new sample ids at the end
+        this.orderedSampleList = new ArrayList<>(savedOrderedSampleList);
+        List<Integer> newSampleIds = this.fileOrderedSampleList.stream().filter(sampleId -> !savedOrderedSampleList.contains(sampleId)).toList();
```
sheridancbio, line 43: I think this is order(N^2) time complexity. For lists containing only a thousand or two samples, this may not be a factor, but at some point (millions? tens of millions?) it might become slow. A faster approach would be to first sort each list and then scan each list in parallel to find the diff.
I suppose I am assuming that contains() is an order(N) operation. If the list is also represented with a hashmap for faster contains() execution then perhaps this comment is unsound.

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
```diff
@@ -0,0 +1,105 @@
...
+        this.orderedSampleList.addAll(newSampleIds);
+        DaoGeneticProfileSamples.deleteAllSamplesInGeneticProfile(this.geneticProfileId);
+        this.storeOrderedSampleList();
```
sheridancbio, line 46: same comment as above (database operations in constructor)

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
```diff
@@ -0,0 +1,105 @@
...
+                throw new IllegalStateException("Number of samples ("
+                        + sampleToValue.size() + ") for genetic entity with id "
+                        + geneticEntityId + " does not match with the number in the inital sample list ("
+                        + initialOrderSampleListSize + ").");
+            }
+        });
+        // add all new sample ids at the end
+        this.orderedSampleList = new ArrayList<>(savedOrderedSampleList);
+        List<Integer> newSampleIds = this.fileOrderedSampleList.stream().filter(sampleId -> !savedOrderedSampleList.contains(sampleId)).toList();
+        this.orderedSampleList.addAll(newSampleIds);
```
sheridancbio, line 44: Here is where the new ordered sample list appends the novel sample ids

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
```diff
@@ -0,0 +1,105 @@
...
+        DaoGeneticProfileSamples.deleteAllSamplesInGeneticProfile(this.geneticProfileId);
+        this.storeOrderedSampleList();
+        daoGeneticAlteration.deleteAllRecordsInGeneticProfile(this.geneticProfileId);
```
sheridancbio, line 47: And here we are wiping the previous genetic profile values, which will be a problem if this is not inside a managed transaction (I am now assuming we are inside a transaction)

sheridancbio (commented):
partial review (Part 3) java classes

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationImporter.java
forus: > Could this function call be moved inside of the store() methods below?
we should call `storeOrderedSampleList()` once only. It's possible to have a flag and do this operation on first call of the `store()` method.

> if the storage of the sample id list succeeds here, but the storage of the cna event profile fails during store() would we have a problem? (or maybe transactions will take care of that?)
this code should run in a single transaction indeed.

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
forus: indeed, this code has to run within single transaction

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
forus: "preexisting" sounds good https://github.com/cBioPortal/cbioportal-core/pull/48/commits/89d6a97bf924c41e00fb9dfab6d7fbd8a01a0f6b

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationImporter.java
forus: code to import genetic entity profiles was repeated in multiple importers. I moved it here.

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationImporter.java
forus: how about having an explicit `initialise()` method that the client side of the code has to call?  https://github.com/cBioPortal/cbioportal-core/pull/48/commits/0edaec3fc67d580dd7ebd30913c50f8ea989de0f

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
forus: addressed in the single commit with the above one https://github.com/cBioPortal/cbioportal-core/pull/48/commits/0edaec3fc67d580dd7ebd30913c50f8ea989de0f

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
forus: Here it is https://github.com/cBioPortal/cbioportal-core/pull/48/commits/9a37bbecf841e15503fddedb55bb48605c552378

src/main/java/org/mskcc/cbio/portal/scripts/ImportExtendedMutationData.java
```diff
@@ -417,9 +457,6 @@ public void importData() throws IOException, DaoException {
                         } else {
                             mutations.put(mutation,mutation);
                         }
-                        if(!sampleSet.contains(sample.getStableId())) {
```
sheridancbio, line 420: This is now folded into the upsert below

sheridancbio (commented):
partial review (part 4) still working through scripts package - up to ImportGeneData

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public boolean store(int geneticEntityId, String[] values) throws DaoException {
+        String[] expandedValues = extendValues(geneticEntityId, values);
+        return super.store(geneticEntityId, expandedValues);
+    }
+
+    @Override
+    public void initialise() {
```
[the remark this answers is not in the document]
forus: ha, the `finalize()` method overrides the method in the `Object` class. I find it appropriate. What do you think?

src/main/java/org/mskcc/cbio/portal/scripts/ImportCopyNumberSegmentData.java
```diff
@@ -66,8 +85,7 @@ private void importData(File file, int cancerStudyId) throws IOException, DaoExc
                     System.err.println("wrong format: "+line);
                 }
 
-                CancerStudy cancerStudy = DaoCancerStudy.getCancerStudyByInternalId(cancerStudyId);
```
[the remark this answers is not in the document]
forus: IDE helped ;)

src/main/java/org/mskcc/cbio/portal/scripts/ImportDrugs.java
```diff
@@ -32,11 +32,18 @@
...
+import org.mskcc.cbio.portal.util.ProgressMonitor;
+import org.mskcc.cbio.portal.util.TsvUtil;
 
-import java.io.*;
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
 
 /**
```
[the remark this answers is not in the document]
forus: Indeed, this class does not seem to do anything. I've created a ticket here https://github.com/cBioPortal/cbioportal-core/issues/54

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
```diff
@@ -0,0 +1,114 @@
...
+    public void finalise() {
+        expandRemainingGeneticEntityTabDelimitedRowsWithBlankValue();
+        super.finalise();
+    }
+
+    private String[] extendValues(int geneticEntityId, String[] values) {
+        Map<Integer, String> sampleIdToValue = mapWithFileOrderedSampleList(values);
+        String[] updatedSampleValues = new String[orderedSampleList.size()];
+        for (int i = 0; i < orderedSampleList.size(); i++) {
+            updatedSampleValues[i] = "";
```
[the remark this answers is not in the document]
forus: I am not sure what action points are here

scripts/importer/validateData.py
sheridancbio: Ok. (although this would require the user to run the validator, unless a validator run occurs as part of scripted study import automatically)

src/main/java/org/mskcc/cbio/portal/dao/DaoClinicalData.java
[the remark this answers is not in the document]
sheridancbio: 👍

src/main/java/org/mskcc/cbio/portal/dao/DaoCnaEvent.java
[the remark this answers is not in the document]
sheridancbio: 👍

src/main/java/org/mskcc/cbio/portal/dao/DaoCopyNumberSegment.java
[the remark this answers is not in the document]
sheridancbio: 👍

src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticAlteration.java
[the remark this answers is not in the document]
sheridancbio: I'm not sure what the right answer is. I think it may be good for the cBioPortal project as a whole to choose and document a standard for comma-separated data value formatting and that all such values be validated and rejected if they conflict with the standard. In general I think it is a worse option to try to accept / accommodate two different flavors of format. If I needed to set the standard I think my personal preference would be to interpret a trailing comma to always indicate that there was another null value at the end of the list (after the final comma). So perhaps some values would be considered errors under that standard. But this is only my opinion/style preference. So I think the topic can be followed up on later.

src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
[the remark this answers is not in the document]
sheridancbio: My memory is that there are two separate DaoCancerStudy deleteCancerStudy functions ... one which relies on the cascade of foreign keys and one which explicitly deletes all data types with a separate dao delete function call for each data type / involved table. But if there are no calls to this function in the cbioportal-core codebase then I think deleting it is fine. (I don't think it will break the MSK importer, but if so, we can work around it)

src/main/java/org/mskcc/cbio/portal/dao/DaoSampleList.java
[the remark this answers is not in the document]
sheridancbio: 👍

src/main/java/org/mskcc/cbio/portal/dao/DaoSampleList.java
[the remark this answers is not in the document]
sheridancbio: 👍

src/main/java/org/mskcc/cbio/portal/dao/DaoSampleList.java
sheridancbio: No .. no problems. My comment was only documenting my understanding during my read through.

src/main/java/org/mskcc/cbio/portal/dao/DaoSampleProfile.java
sheridancbio: 👍

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationImporter.java
sheridancbio: Yes, the initialise (english spelling) call is good I think. Sometimes in cases like this, rather than depending on the caller/client code to call in to initialize() before using the functionality, there is a data member which can be used to determine whether or not the initialization has occurred yet or not. Then there can be a conditional initialization ... a test in the actual business logic which asks (is this initialized yet? ... if not yet then call initialize()). That way the initialization cannot be erroneously skipped in the client code.

But this looks good to me.

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
sheridancbio: 👍

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
sheridancbio: Ok, that solution is even better. The set lookup can be (in theory) order 1 which give better overall complexity. Thanks.
👍

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
sheridancbio: 👍

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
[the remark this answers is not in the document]
sheridancbio: Anything is ok ... I'm kind of sorry for bringing up the british/american dialect topic in the first place.

src/main/java/org/mskcc/cbio/portal/scripts/GeneticAlterationIncrementalImporter.java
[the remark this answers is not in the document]
sheridancbio: I guess my understanding of this topic (what kinds of metrics are acceptable in the genetic_alteration table) is simply inadequate. I think it has been expanded beyond the CNA event calls, so I would not feel comfortable making any decisions about how best to validate or import into that table. There is some guidance in the file format documentation about unset values, but I don't think it is clear enough. There also seems (to me) to be two places where we define the set of samples in a profile (at least for CNA profiles) ... we have the list of columns across the top headers of the file which give the sample list, and we also have case_lists/cases_cna.txt. I'm not sure which one should "rule" when it comes to data import .. or what should be done if there is a clash. I think it would be better to consolidate the definition of the sample list into one location .. or to require that both lists match exactly if that is not feasible. But I don't have any action points for this PR I guess.

src/main/java/org/mskcc/cbio/portal/scripts/ImportCnaDiscreteLongData.java
```diff
@@ -22,63 +22,116 @@
...
-        this.cnaFile = cnaFile;
-        this.geneticProfileId = geneticProfileId;
-        this.genePanel = genePanel;
-        this.daoGene = daoGene;
-        this.geneticAlterationGeneImporter = new GeneticAlterationImporter(geneticProfileId, daoGeneticAlteration);
+       this(cnaFile, geneticProfileId, genePanel, daoGene, namespaces, false);
+    }
+    public void importData() {
+        JdbcUtil.getTransactionTemplate().execute(status -> {
+            try {
```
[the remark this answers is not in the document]
sheridancbio: 👍

src/main/java/org/mskcc/cbio/portal/scripts/ImportDrugs.java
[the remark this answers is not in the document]
sheridancbio: Thanks

sheridancbio (approved):
This looks good. Legacy functionality is supported and new functionality is enabled through command line arguments. Changes are described in the approved RFC-79 document and code follows that closely.

The remainder of the java code (after my part 4 partial review) was reviewed by @forus and I interactively yesterday. My review did not cover the test case data, but testing approaches (including python script "called" java methods) was discussed as well.

The MSK importer has not yet been tested with these changes, but I believe that any problems which are observed can be addressed in the MSK codebase itself (which mainly wraps/depends on the DAO classes here)

Issue #36, opened 2024-05-01
Move integration tests from cbioportal to cbioportal-core


forus:
Since we got cbioportal-core project dockerised, we can move integration tests that load data to cbioportal-core more easily.

https://github.com/cBioPortal/cbioportal/blob/master/.github/workflows/integration-test.yml#L67

PR #55, opened 2024-07-19
Initialise transactionTemplate if needed

forus:
`JdbcUtil.getTransactionTemplate()` returned `null` causing null pointer exception for all commands that used transaction programatically:
- `RemoveSamples`
- `RemovePatients`
- `ImportCnaDiscreteLongData`
- `ImportTabDelimData`

### DIFF
--- a/src/main/java/org/mskcc/cbio/portal/dao/JdbcUtil.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/JdbcUtil.java
@@ -82,6 +82,9 @@ public class JdbcUtil {
     }
 
     public static TransactionTemplate getTransactionTemplate() {
+        if (transactionTemplate == null) {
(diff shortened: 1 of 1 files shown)


sheridancbio (approved):

pieterlukasse (commented):
I think it would be good to include a test covering `getTransactionTemplate` in this case. Apparently the existing set of tests were not sufficient for catching this in CI 🤔

forus:
> I think it would be good to include a test covering `getTransactionTemplate` in this case. Apparently the existing set of tests were not sufficient for catching this in CI 🤔

@pieterlukasse this case is covered by tests in cbiportal https://github.com/cBioPortal/cbioportal-core/issues/36

pieterlukasse (approved):

PR #58, opened 2024-09-05
Minor fix to validate floats in clinical files

rmadupuri:
The current check `if (value[0] in ('>','<')) and value[1:].isdigit():` throws an error if the values are floats. This PR fixes that.


MatthijsPon (approved):
Good catch!

PR #59, opened 2024-09-27
Clickhouse import process

sheridancbio:
Scripts, property files, and documentation for the blue-green deployment strategy to enable MySQL/Clickhouse database updates during cancer study import.


scripts/clickhouse_import_support/copy_mysql_database_tables_to_clickhouse.sh
```diff
@@ -0,0 +1,263 @@
...
+        return 1
+    fi
+    if ! parse_property_file "$properties_filepath" my_properties ; then
+        usage
+        return 1
+    fi
+    if [ "$database_to_transfer" == "blue" ] ; then
+        clickhouse_destination_database_name="${my_properties['clickhouse_blue_database_name']}"
+        mysql_source_database_name="${my_properties['mysql_blue_database_name']}"
+    else
```
mandawilson, line 52: Here we could check explicitly that the database_to_transfer is green or give an error if it isn't blue or green.

scripts/clickhouse_import_support/create_derived_tables_in_clickhouse_database_by_profile.py
```diff
@@ -0,0 +1,78 @@
...
+
+import datetime
+import re
+import subprocess
+
+def get_list_of_genetic_profile_ids():
+    get_genetic_profile_id_list_query="SELECT genetic_profile_id FROM genetic_profile WHERE genetic_alteration_type NOT IN ('GENERIC_ASSAY', 'MUTATION_EXTENDED', 'STRUCTURAL_VARIANT')"
+    query_argument_template="--query={0}"
+    query_argument = query_argument_template.format(get_genetic_profile_id_list_query)
+    clickhouse_client_obtain_genetic_profile_id_list = ["clickhouse", "client", "--config-file=clickhouse_client_config_2024-10-14-09-03-02.yaml", query_argument]
```
mandawilson, line 11: Should the config-file always hardcoded to this?

mandawilson: I see below you have a TODO to deal with that.

scripts/clickhouse_import_support/derived_table_construction_commands.sql
```diff
@@ -0,0 +1,36 @@
+DROP TABLE IF EXISTS sample_to_gene_panel_derived;
+DROP TABLE IF EXISTS gene_panel_to_gene_derived;
```
mandawilson, line 2: Aren't these commands stored somewhere else?

scripts/clickhouse_import_support/get_database_currently_in_production.sh
```diff
@@ -0,0 +1,110 @@
...
+    if ! parse_property_file "$properties_filepath" my_properties ; then
+        usage
+        return 1
+    fi
+    if ! initialize_mysql_command_line_functions ; then
+        usage
+        return 1
+    fi
+    update_management_database_name="${my_properties['mysql_update_management_database']}"
+    ### TODO : fix this
```
mandawilson, line 42: I think this needs adding back.

scripts/clickhouse_import_support/set_update_process_state.sh
```diff
@@ -0,0 +1,138 @@
...
+    if ! parse_property_file "$properties_filepath" my_properties ; then
+        usage
+        return 1
+    fi
+    if ! initialize_mysql_command_line_functions ; then
+        usage
+        return 1
+    fi
+    update_management_database_name="${my_properties['mysql_update_management_database']}"
+    ### TODO : fix this
```
mandawilson, line 41: Add back?

mandawilson (dismissed):

scripts/clickhouse_import_support/derived_table_construction_commands.sql
sheridancbio: Yes. This file was a placeholder. I need to add calls to the script which processes the .sql files from the github repo (clickhouse.sql and materialized_views.sql) and bursts them into individual sql statements (1 per file). Also, we need to detect and skip over the special cases (genetic_alteration_derived, generic_assay_data_derived).

So this file will be going away

scripts/clickhouse_import_support/copy_mysql_database_tables_to_clickhouse.sh
sheridancbio: We do an explicit "is it other than 'green' or 'blue'" check above, so that is not possible

scripts/clickhouse_import_support/set_update_process_state.sh
sheridancbio: yes

scripts/clickhouse_import_support/clickhouse_client_command_line_functions.sh
```diff
@@ -0,0 +1,229 @@
...
+unset configured_clickhouse_config_file_path
+unset sql_data_field_value
+unset sql_data_array
+configured_clickhouse_config_file_path=""
+sql_data_field_value=""
+declare -a sql_data_array
+clickhouse_client_database_exists_filepath="$(pwd)/ccclf_database_exists.txt"
+clickhouse_client_database_table_list_filepath="$(pwd)/ccclf_database_table_list.txt"
+
+function write_clickhouse_config_file() {
```
alisman, line 12: @sheridancbio it would be nice if these config files existed and just used env variables (not sure if it's possible).   alternatively, i'm pretty sure that the CLI commands accept arguments.  so, rather than creating a file, would it not be better just to call the command with the arguments?

sheridancbio: The motivation behind avoiding command line arguments which contain secrets and credentials is that on a shared unix system, any user on the system usually has access to the executed command (with arguments) and the bash environment settings of all commands running on the system. So having a credential (like a mysql password) in a command line argument or in an environment variable exposes that credential to everyone who is logged in to the unix host. Because all of the standard tools which we are using can do argument reading from config files, and because the config files can be made readable only by the running process, it is more secure to pass credentials into these update processes through a (protected) configuration file.  The sling tool also kind of requires using a config file to construct a set of sling data "connections" which are referred to with connection references.

But when using a docker image deployment (where the whole system is basically single-user) these security concerns are less important. I think @zainasir does plan to leverage environment variables for the passing in of configuration to the docker container. These environment variables can be referenced inside the container to output a functional properties file that the tools can use.

zainasir (approved):
LGTM!

PR #62, opened 2024-10-09
Removing the filtering of amino acid changes with '+' as not relevant anymore

MatthijsPon:
This change was not taken along when the externalized core was created. See https://github.com/cBioPortal/cbioportal/commit/76fd01b773fbd976a085914943f54fb89877c4df for the previous time this was implemented


rmadupuri (approved):
Good catch!

PR #63, opened 2024-10-15
Validation Update: Add a warning for GERMLINE variants in MAF and SV files

rmadupuri:
This PR adds a validation check to give a warning message when GERMLINE variants are present in MAF and SV tables to prevent the accidental release of these variants to the public.


MatthijsPon (approved):
Looks good! Thanks for also updating the tests

PR #70, opened 2024-12-12
Fix false negatives for validator for GDC data

jamesqo:
See https://github.com/cBioPortal/datahub-study-curation-tools/pull/65 for an explanation. The copy of the validator used in this repo is used to validate the studies in Datahub. The false negatives it's reporting are currently preventing the TARGET studies from being merged.


sheridancbio (approved):
merged for test conditions pass

PR #75, opened 2025-02-20
update tool in github action for mvn build

sheridancbio:
attempt to fix failing action step


sheridancbio:
Updating the version of advanced-security third party tool did not reliably fix the failures. The upload of the dependency tree seems to succeed sometimes and fail sometimes (under either version of the tool). Because the tool behavior is inconsistent, I believe we should not make a successful completion of this step in the git action required. If the attempt fails, the automated checks could still be made to pass.

inodb (approved):
Sounds reasonable 👍

Any clue why it fails occasionally? Is it a permission issue (e.g. can't do it from fork) or kinda random?

inodb (approved):

PR #76, opened 2025-02-27
 add retry loop for http requests to github

sheridancbio:
- detect when ratelimit has been triggered

This script is no longer intended for use in automated pipeline runs, but may still be useful for building the docker image for the database transfer tools. That build will (in the future) likely contain a number of baked in versions of the derivation scripts for use with various schema versions of cBioPortal databases (so that the latest image will work with previous versions of the database schema as well as the latest database schema).

This script can be used to assemble the various known versions of the clickhouse.sql file and associated files.

The change in this PR introduces 2 things:
1. up to 5 tries will be made for fetches from github through the github content download API
2. the "ratelimit" condition is now detected and reported to the user with intelligible warnings/errors. If the ratelimit reset is approaching (within 5 minutes) the program will wait for the reset to happen before reattempting download.

### DIFF
--- a/scripts/clickhouse_import_support/download_clickhouse_sql_scripts_py3.py
+++ b/scripts/clickhouse_import_support/download_clickhouse_sql_scripts_py3.py
@@ -6,10 +6,15 @@ import base64
 import http.client
 import json
 import os
+import time
 import sys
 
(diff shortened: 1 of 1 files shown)


zainasir (approved):
Looks good to me! Thanks for the fix.

Issue #74, opened 2025-02-17
Incremental update not allowed on CNA_DISCRETE_LONG datatype


flelayo:
In the [documentation](https://docs.cbioportal.org/incremental-data-loading/), the CNA_DISCRETE_LONG datatype is supported but when running metaImport.py in incremental mode it did not work with this error message:
`This type does not support incremental upload: meta_CNA_long`

I added the meta_CNA_long to the supported type in the cbioportal_common.py and it seems the import went well and the CNV value where added to the study.

```diff
# order is important! This is the order in which they should be loaded:
INCREMENTAL_UPLOAD_SUPPORTED_META_TYPES = [
    MetaFileTypes.PATIENT_ATTRIBUTES,
    MetaFileTypes.SAMPLE_ATTRIBUTES,
    MetaFileTypes.MUTATION,
    MetaFileTypes.MUTATION_UNCALLED,
    MetaFileTypes.EXPRESSION,
    MetaFileTypes.CNA_DISCRETE,
+   MetaFileTypes.CNA_DISCRETE_LONG,
    MetaFileTypes.CNA_CONTINUOUS,
    MetaFileTypes.CNA_LOG2,
    MetaFileTypes.METHYLATION,
    MetaFileTypes.PROTEIN,
    MetaFileTypes.GENERIC_ASSAY_CONTINUOUS,
    MetaFileTypes.GENERIC_ASSAY_BINARY,
    MetaFileTypes.GENERIC_ASSAY_CATEGORICAL,
    MetaFileTypes.TIMELINE,
    MetaFileTypes.GENE_PANEL_MATRIX,
    MetaFileTypes.STRUCTURAL_VARIANT,
    MetaFileTypes.SEG,
]
```

Can you assure me CNA_DISCRETE_LONG is supported ?

Thanks.

PR #92, opened 2025-09-26
Add Cna discrete long incremental

TJMKuijpers:
The validator does not allow to load CNA discrete long format: https://github.com/cBioPortal/cbioportal-core/issues/74 . 

When this type is added to `INCREMENTAL_UPLOAD_SUPPORTED_META_TYPES` the validator does allow CNA discrete long format and the loader does continue to load the new data.


TJMKuijpers:
@forus I'm still working on the tests but I'm not sure what the current tests do: they check if the call is being made and returns a success status?

PR #103, opened 2026-01-13
Remove cbioportal dependency

forus:
- cbioportal class references (in package org.cbioportal -- mainly models) were:
        - removed where not needed
        - remapped to local classes where possible
            for example: CNA enum references mapped to matching CopyNumberStatus static constants
        - otherwise duplicated into cbioportal-core. Examples:
            - GenericEntityProperty model class
            - EntityType model class
            - GeneticEntity model class
            - GenericAssayMeta model class
            - ResourceType model class
    - duplicated classes located in package: org.mskcc.cbio.portal.model.shared
    - implemented other functionality via new DAO functionality
            for example: dump portal info task (with new tests)
    - standardized import statements (avoiding wildcard except for java.io/java.util)
    - drop spring contect references/resources which are not needed for import
    - during test execution, logic requesting user input for certain updates was disabled by
        throwing an exception, unless this argument is provided:
            --confirm-delete-all-genesets-hierarchy-genesetprofiles
    - relocated TsvUtil tests relocated to parallel package
    - removed cbioportal artifact dependency, and directly added several dependencies which were
        previoulsy provided transitively through the cbioportal artifact.

### DIFF
--- a/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
+++ b/src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
@@ -33,7 +33,10 @@
 package org.mskcc.cbio.portal.util;
 
 import java.io.*;
+import java.nio.charset.StandardCharsets;
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -44,18 +47,20 @@ public class GlobalProperties {
 
     public static final String HOME_DIR = "PORTAL_HOME";
     private static final String PORTAL_PROPERTIES_FILE_NAME = "application.properties";
+    private static final String POM_RESOURCE_PATH = "META-INF/maven/org.mskcc.cbio/core/pom.xml";
+    private static final String DEFAULT_DB_VERSION = "0";
+    private static final Pattern DB_VERSION_PATTERN =
+        Pattern.compile("<db\\.version>\\s*([^<]+)\\s*</db\\.version>");
 
-    public static final String APP_VERSION = "app.version";
     public static final String DB_VERSION = "db.version";
     public static final String SPECIES = "species";
     public static final String DEFAULT_SPECIES = "human";
-    public static final String NCBI_BUILD = "ncbi.build";
-    public static final String DEFAULT_NCBI_BUILD = "37";
     public static final String UCSC_BUILD = "ucsc.build";
     public static final String DEFAULT_UCSC_BUILD = "hg19";
 
     private static Logger LOG = LoggerFactory.getLogger(GlobalProperties.class);
     private static ConfigPropertyResolver portalProperties = new ConfigPropertyResolver();
+    private static volatile String cachedDbVersion;
 
     /**
      * Minimal portal property resolver that takes system property overrides.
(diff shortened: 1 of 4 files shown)


forus:
> I contributed to this PR, so I thought that I should just leave a comment. My one concern is around the command line argument `--confirm-delete-all-genesets-hierarchy-genesetprofiles` I see that we are providing this argument during the running of the tests (I think) ... but I don't see how it will be affecting the running of the tests (maybe I just missed that). I think we should have a discussion about the interactive mode which was there for the updating of gene sets and gene panels. I have some ideas of how to make this safer to do, or to avoid the alteration (which could have side effects on any loaded study which has events involving modified genes)
> 
> We should decide if we want to keep this interactive "are you sure" mode or not ... and if we want to keep it we should test it to make sure it still works.

@sheridancbio 
> but I don't see how it will be affecting the running of the tests

The re-uploading of the geneset in the `TestIntegrationTest` should not affect followup tests as there is `@Rollback` annotation.
It will become a problem though for tests relying or rolling back changes after we convert application to ClickHouse as it does not have transaction. Something to think on.

> I have some ideas of how to make this safer to do, or to avoid the alteration (which could have side effects on any loaded study which has events involving modified genes)

maybe you can write down your ideas on a separate ticket? Avoiding the alteration in the first place should solve the above issue by removing reliance on rolling back changes during testing.

forus:
@sheridancbio I just realized something. While removing the cbioportal dependency, we changed how cbioportal-core learns which database schema version it supports, and what the default species and genome build are.

Before, cbioportal-core read the supported database version from the mvn.properties file inside the cbioportal JAR. We changed this so that the property is now in cbioportal-core’s own application.properties file instead. It also seemed to consider the application.properties inside the cbioportal JAR file to read the default species and genome build. Right?

We need to better understand what this change means for unpacking cbioportal-core inside the cbioportal container. With the current setup, it will not work as it does today. How the unpacked version used to get these inputs previously and how it'd work now?

See this Dockerfile line for context:
https://github.com/cBioPortal/cbioportal/blob/master/docker/web-and-data/Dockerfile#L47

I think we should understand this impact better before merging into master. What do you think?

cc @n1zea144 @pieterlukasse

src/main/java/org/mskcc/cbio/portal/util/GlobalProperties.java
```diff
@@ -169,12 +215,21 @@ public static String getSpecies(){
     	}
 
     public static String getDbVersion() {
-        String version = portalProperties.getProperty(DB_VERSION);
-        if (version == null)
-        {
-            return "0";
+        String override = System.getProperty(DB_VERSION);
```
sheridancbio: Does System.getProperty() also detect environment variable settings?  If we are allowing the setting of DB_VERSION with java -DDB_VERSION=99.99.99 then we should probably also support 'export DB_VERSION="99.99.99" ; java .... '

When launching in a container (such as when managed by kubernetes or when creating a published docker image), it is sometimes easier to initialize properties as environment variables. So I recommend adding something like a check of System.getenv() too if that is not already handled as part of System.getProperty()

sheridancbio (commented):
I've looked at the added changeset which switched away from maven.properties for DB_VERSION, and instead parses the pom.xml file to find the "built in" version string.

This looks good.

I suggested adding the override mechanism of an environment variable as well (in addition to the java property way).

sheridancbio:
@pieterlukasse We plan to add comparison/compatibility checks that at a minimum, the classes which were dependencies (and are now located in package org.mskcc.cbio.portal.model.shared) are compatible with their analogs in repository cbioportal. But there are additional classes in cbioportal-core which should be kept compatible with corresponding classes in cbioportal, so a more general effort to identify and check compatibility between repositories makes sense.

PR #107, opened 2026-01-21
Convert MySQL to ClickHouse SQL dialect + Implement ID auto-increment in CH database

forus:
import directly to Clickhouse database
- Dao layer SQL converted from MySQL compatibility into Clickhouse compatibility
- internal id auto-increment class with on-exit persistence in database table


n1zea144 (dismissed):
Minor comments, mostly to make sure I understand what is happening.  Nice job and cleanup when you had the opportunity.

sheridancbio (commented):
This comment is focussed on the latest changeset added to the PR (introducing a change to the incremental update of genetic_alteration table entries)

I've not yet fully understood the incremental update logic that this PR is based on. At first, the main change here seems to be a switch to a ReplacingMergeTree engine so we can do efficient alteration on a per-record basis. Previously we (maybe) were dropping all records for the samples of the profile from the genetic_alteration table, and then recreating them with the longer comma separated value lists. Now we are overwriting on a per record bases as we make the update.

My first general suggestion is that rather than relying on the user to notice the failure and take corrective action (re-uploading the full study), we should either:
1. do what we can ourselves to correct the data state --- which may be difficult if, for example, the incremental update involved deleting a sample --- in that case, cna event values may be deleted for some of the genetic_alteration records. Going "back" to the previous state would require reintroduction of the lost values. If we did not save them as we were deleting them we would not have an easy way to reintroduce them.
2. remedy the error by deleting the entire study from the database in order to return it to a consistent state. I think it is safer to remove a corrupted study from the database rather than tolerate the presence of erroneous data. A slightly less disruptive correction would be to delete the genetic profile which is connected to the corrupted data in genetic_alteration. But perhaps that should be avoided because it would give the user the impression that no CNA profiling was done on the samples ... when in fact the profiling *was* done. Also, we would need to adjust the sample_profile table to remove the cna gene panel references for the study too. Probably better to remove the corrupted study and force the user to do a full-study import in order to return the study into their cBioPortal deployment.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseAutoIncrement.java
```diff
@@ -0,0 +1,220 @@
...
+        synchronized (ClickHouseAutoIncrement.class) {
+            if (SEQUENCE_TABLE_READY.get()) {
+                return;
+            }
+            Connection con = null;
+            PreparedStatement stmt = null;
+            try {
+                con = JdbcUtil.getDbConnection(ClickHouseAutoIncrement.class);
+                stmt = con.prepareStatement(
+                    "CREATE TABLE IF NOT EXISTS " + SEQUENCE_TABLE + " ("
```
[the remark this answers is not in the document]
sheridancbio: when we implement outer green-blue operations in clickhouse we will need to think about whether we clone the derived tables or not ... perhaps they are going to be completely overwritten anyway.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkLoader.java
```diff
@@ -0,0 +1,248 @@
...
+            con = JdbcUtil.getDbConnection(ClickHouseBulkLoader.class);
+            List<String> columns = resolveColumnNames(con);
+            validateRecordWidths(columns.size());
+
+            byte[] payload = buildTsvPayload(columns);
+            stmt = con.prepareStatement(buildInsertStatement(columns));
+            stmt.setBinaryStream(1, new ByteArrayInputStream(payload));
+
+            int rowsInserted = stmt.executeUpdate();
+            if (rowsInserted <= 0) {
```
sheridancbio: I wonder what the rationale is for this

src/main/java/org/mskcc/cbio/portal/dao/DaoCancerStudy.java
```diff
@@ -629,23 +600,65 @@ public static void purgeUnreferencedRecordsAfterDeletionOfStudy() throws DaoExce
...
+        } finally {
+            JdbcUtil.closeAll(DaoCancerStudy.class, null, pstmt, rs);
+        }
+        return ids;
+    }
+
+    private static void deleteByStudyId(Connection con, String sql, int cancerStudyId) throws SQLException {
+        try (PreparedStatement pstmt = con.prepareStatement(sql)) {
+            pstmt.setInt(1, cancerStudyId);
+            pstmt.executeUpdate();
```
sheridancbio: The original code didn't bother to check the return value from calling executeUpdate() .... that's too bad. We don't have to improve on the original code I guess .... but I think that was an oversight when this code was added. At the very least I would have suggested logging a warning if the update to the cancer_study table resulted in a "0 records altered" result.

src/main/java/org/mskcc/cbio/portal/dao/DaoClinicalData.java
```diff
@@ -43,6 +41,20 @@
...
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
```
sheridancbio: explicit imports of classes in these standard common packages is probably not helpful. But if we want to adopt a clear absolute rule of "no wildcard imports" I would be ok with that. I just urge consistency.

src/test/resources/seed_mini.sql
```diff
@@ -1,684 +1,553 @@
...
---
--- This program is distributed in the hope that it will be useful,
--- but WITHOUT ANY WARRANTY; without even the implied warranty of
--- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
--- GNU Affero General Public License for more details.
---
--- You should have received a copy of the GNU Affero General Public License
--- along with this program.  If not,see <http://www.gnu.org/licenses/>.
--- ----------------------------------------------------------------------------
--- A manually extracted subset of data for a small number of genes and samples from the BRCA
```
sheridancbio: a comment of where this came from (BRCA) and the original developer who extracted this data seems helpful. Maybe change the comment to "Based on ...." because there have been extensive additions.

src/test/java/org/mskcc/cbio/portal/integrationTest/IntegrationTestBase.java
```diff
@@ -13,56 +13,74 @@
...
+    private static final String CLICKHOUSE_IMAGE = "clickhouse/clickhouse-server:24.3";
     private static final String SKIP_ENV = "CBIOPORTAL_TEST_DB_SKIP";
     private static final String TARGET_DIR = "target/test-db";
     private static final String VERSION_FILE = "cbioportal.version";
-    private static final String CGDS_SQL_FILE = "cgds.sql";
+    private static final String CGDS_SQL_FILE = "clickhouse_cgds.sql";
+    private static final String SEED_SQL_FILE = "seed_mini.sql";
     private static final String CGDS_URL_TEMPLATE =
-        "https://raw.githubusercontent.com/cBioPortal/cbioportal/%s/src/main/resources/db-scripts/cgds.sql";
+        "https://raw.githubusercontent.com/cBioPortal/cbioportal/%s/src/test/resources/clickhouse_cgds.sql";
```
sheridancbio: 👍

src/test/java/org/mskcc/cbio/portal/integrationTest/scripts/TestImportClinicalData.java
```diff
@@ -423,24 +420,22 @@ public void testImportClinicalData_PatientIdError() throws Exception {
...
         // set the info usually parsed from args
         importClinicalData.setFile(cancerStudy, clinicalFile, type + "_ATTRIBUTES", false);
         importClinicalData.importData();
         // loading twice should also give error
-        exception.expect(DaoException.class);
-        // it is not a specific "duplication" error message but a general DB error since the 
-        // validation only gives specific error when in same file (maybe at some point we want to support clinical data 
-        // in multiple PATIENT and SAMPLE files(?) ):
-        exception.expectMessage("DB Error");
+        exception.expect(RuntimeException.class);
```
sheridancbio: potentially, an unexpected RuntimeException which is not from the DaoClasses might inadvertently cause this test to "pass". Maybe we can discuss why this changed to RuntimeException.

src/test/java/org/mskcc/cbio/portal/integrationTest/scripts/TestDumpPortalInfoApproval.java
```diff
@@ -68,7 +68,7 @@ private static void assertApprovedFilesMatch(Path receivedDir, Path approvedDir)
     private static void assertJsonMatch(String fileName, Path approved, Path received) throws JSONException, IOException {
         String message = "Mismatch for " + fileName + ".\nCopy " + received.toAbsolutePath()
                 + " to " + approved.toAbsolutePath() + " after review.";
-        JSONAssert.assertEquals(message, Files.readString(approved), Files.readString(received), true);
+        JSONAssert.assertEquals(message, Files.readString(approved), Files.readString(received), false);
```
sheridancbio: I don't understand this change

sheridancbio (commented):
Partial review, including most of main code updates and all of test code updates.

src/main/java/org/mskcc/cbio/portal/dao/DaoSample.java
```diff
@@ -289,32 +297,35 @@ public static void deleteSamples(int internalStudyId, Set<String> sampleStableId
...
+            deleteByIds(con, "DELETE FROM sample_cna_event WHERE sample_id IN ", internalSampleIds);
+            deleteByIds(con, "DELETE FROM mutation WHERE sample_id IN ", internalSampleIds);
+            deleteByIds(con, "DELETE FROM structural_variant WHERE sample_id IN ", internalSampleIds);
+            deleteByIds(con, "DELETE FROM sample_profile WHERE sample_id IN ", internalSampleIds);
+            deleteByIds(con, "DELETE FROM sample_list_list WHERE sample_id IN ", internalSampleIds);
+            deleteByIds(con, "DELETE FROM copy_number_seg WHERE sample_id IN ", internalSampleIds);
+            deleteByIds(con, "DELETE FROM allele_specific_copy_number WHERE sample_id IN ", internalSampleIds);
+            deleteByIds(con, "DELETE FROM clinical_sample WHERE internal_id IN ", internalSampleIds);
+            deleteByIds(con, "DELETE FROM resource_sample WHERE internal_id IN ", internalSampleIds);
+            deleteByIds(con, "DELETE FROM sample WHERE internal_id IN ", internalSampleIds);
```
sheridancbio: Do we need to add logic here? When a sample is deleted and the patient connected to that sample no longer has any samples remaining in the database shouldn't we delete the patient too? We generally don't allow patient records where the patient has no samples defined. I'm not sure if that is a constraint or merely a convention, but I think we should clarify.

Maybe this case does not occur because maybe the incremental update does not do pure deletion? I think you said that there was a separate "delete incrementally" function. In that case we can ask this question about that codebase.

sheridancbio (commented):
partial review : later Dao classes

src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -48,18 +46,34 @@
 import org.mskcc.cbio.portal.model.Sample;
 import org.mskcc.cbio.portal.util.MutationKeywordUtils;
 
+import java.sql.Connection;
```
sheridancbio: In other classes we often allow wildcard for: java.util and java.sql classes
It is ok to be explicit. I think I have already made a comment in this PR that I suggest consistency and would be ok adopting a "never use wildcard imports" rule

sheridancbio (changes requested):
part 3 (final part .. review is now complete at this time)

Across this review I've requested a few changes and made several suggestions. Some additional use case testings may be needed based on this (part 3) review.

PR #108, opened 2026-01-28
feat: Add duplicate SAMPLE_ID validation for timeline files (cBioPort…

hetanshee:
Ref cBioPortal/cbioportal#10917

## Description

- Add duplicate SAMPLE_ID validation for timeline files
- Warn when SAMPLE_IDs appear multiple times in SPECIMEN, SAMPLE ACQUISITION, or SEQUENCING events
- Per cBioPortal docs, these event types render as numbered sample discs, duplicates cause conflicting visualization.

### DIFF
--- a/scripts/importer/validateData.py
+++ b/scripts/importer/validateData.py
@@ -3491,10 +3491,27 @@ class TimelineValidator(Validator):
         'EVENT_TYPE']
     REQUIRE_COLUMN_ORDER = True
     ALLOW_BLANKS = True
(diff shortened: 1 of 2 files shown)


hetanshee:
I took some time to revisit this and wanted to check my understanding.

My understanding is that #129 covers full-record duplication, while this PR targets duplicate `SAMPLE_ID`s even when rows are not identical.

I included `SPECIMEN` with `SAMPLE ACQUISITION` and `SEQUENCING` because they seemed closely related in the timeline model, but I’d appreciate guidance on whether that’s the right scope here. 

I’m also leaning toward making this an error instead of a warning.

PR #117, opened 2026-03-04
DRAFT - Add scripts to support ClickHouse flow for cmo-pipelines

jamesqo:


scripts/clickhouse_import_support/manage_cbioportal_databases_tool.properties
```diff
@@ -1,67 +0,0 @@
-# clickhouse import support tools properties file
```
sheridancbio: We need an example properties file to show users how to configure tools on their system. This file should not be deleted.

sheridancbio (changes requested):
Maybe this PR is out of date

PR #126, opened 2026-03-26
A java command to check constraints violoations in CH database

forus:
## Summary

  This PR adds a Java-based ClickHouse constraint check and documents how to run it.

  ### What changed

  - Added `ClickHouseConstraintChecker` to detect ClickHouse foreign-key and unique-key violations for the cBioPortal schema.
  - Added `org.mskcc.cbio.portal.scripts.CheckClickHouseConstraints` as a runnable CLI entrypoint.
  - Kept backward compatibility with the older referential-integrity API through a deprecated `ClickHouseReferentialIntegrityChecker` wrapper.
  - Documented how to run the constraint check from both:
    - the built `core-*.jar`
    - the `cbioportal-core` Docker image

  ### Implementation details

  - Foreign-key violations are detected by generated ClickHouse SQL against the known schema relationships.
  - Unique-key violations are detected by grouping on expected key columns and reporting duplicate rows.
  - The CLI groups and formats violations for console output and exits with a non-zero status when violations are found.
  - The constraint definitions are centralized in code and intended to stay in sync with `clickhouse_cgds.sql`.

  ## Test coverage

  - Added integration tests for foreign-key violation detection.
  - Added integration tests for unique-key violation detection.
  - Added SQL fixtures covering violating rows across the ClickHouse schema.
  - Updated the ClickHouse integration test setup to support this validation flow.

  ## Documentation

  - Added a README section describing how to check ClickHouse constraint violations.
  - Included example commands for both direct JAR execution and Docker-based execution.

  ## Testing

  - Added integration tests:
    - `TestClickHouseConstraintChecker`
  - Added fixtures:
    - `clickhouse_fk_violations.sql`
    - `clickhouse_unique_violations.sql`

  If you want, I can also turn this into a shorter PR body plus a one-line PR title.

### DIFF
--- a/src/main/java/org/mskcc/cbio/portal/dao/ClickHouseConstraintChecker.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/ClickHouseConstraintChecker.java
@@ -1,0 +1,629 @@
+package org.mskcc.cbio.portal.dao;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Generates and executes ClickHouse SQL that reports constraint violations
+ * (foreign keys and unique keys) for the cBioPortal ClickHouse schema.
+ * <p>
+ * Foreign-key output columns per violation row:
+ * referer_table, referer_columns, referer_values, referred_table, referred_columns
+ * <p>
+ * Unique-key output columns per violation row:
+ * table_name, key_columns, key_values, duplicate_count
+ * <p>
+ * Notes:
+ * - ClickHouse does not enforce FKs, so orphans are detected via LEFT JOIN + isNull().
+ * - Unique keys are detected via GROUP BY + HAVING count() > 1.
+ * - Table/column names are normalized to lower-case in the generated SQL.
+ * - Rows are checked only when all FK/unique-key columns are non-NULL.
+ * - "NULL" in referer/key values means the column value was NULL.
+ * <p>
+ * The constraint lists are hard-coded for the ClickHouse schema (see clickhouse_cgds.sql).
+ * Update {@link #schemaForeignKeys()} and {@link #schemaUniqueKeys()} when the schema changes.
+ */
+public class ClickHouseConstraintChecker {
+
+    public static final String REFERER_TABLE = "referer_table";
+    public static final String REFERER_COLUMNS = "referer_columns";
+    public static final String REFERER_VALUES = "referer_values";
+    public static final String REFERRED_TABLE = "referred_table";
+    public static final String REFERRED_COLUMNS = "referred_columns";
+
+    public static final String UNIQUE_TABLE = "table_name";
+    public static final String UNIQUE_COLUMNS = "key_columns";
+    public static final String UNIQUE_VALUES = "key_values";
+    public static final String UNIQUE_DUPLICATE_COUNT = "duplicate_count";
+
+    private static final String VALUE_SEPARATOR = "|";
+
+    private static final List<ForeignKey> SCHEMA_FOREIGN_KEYS = schemaForeignKeys();
+    private static final List<UniqueKey> SCHEMA_UNIQUE_KEYS = schemaUniqueKeys();
+    private static final String FOREIGN_KEY_MASTER_QUERY = generateMasterUnionQuery(
+            SCHEMA_FOREIGN_KEYS,
+            fk -> fk.childTable,
+            fk -> generateForeignKeyViolationSelect(fk, "c", "p")
+    );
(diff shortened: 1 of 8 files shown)


jamesqo (commented):
Looks like really good work! Thanks for tagging me and giving me the opportunity to review

src/test/java/org/mskcc/cbio/portal/integrationTest/dao/TestClickHouseConstraintChecker.java
```diff
@@ -0,0 +1,277 @@
+package org.mskcc.cbio.portal.integrationTest.dao;
```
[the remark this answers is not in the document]
forus: @jamesqo I like your changes. Makes sense!

forus:
@jamesqo @sheridancbio Please review

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseConstraintChecker.java
```diff
@@ -0,0 +1,629 @@
...
+                .map(col -> "isNull(" + tableAlias + "." + quote(col) + ")")
+                .collect(Collectors.joining(" AND ", "(", ")"));
+    }
+
+    /**
+     * Builds the key-values expression, joining values with '|' and rendering NULL as "NULL".
+     */
+    private static String buildKeyValuesExpression(List<String> columns, String tableAlias) {
+        List<String> values = columns.stream()
+                .map(col -> "ifNull(toString(" + tableAlias + "." + quote(col) + "), 'NULL')")
```
sheridancbio, line 582: Will we run into false positives when tables contain fields with value "NULL" in a field? Perhaps we need a placeholder value for NULL which is more likely to not be observed in the actual table data.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseReferentialIntegrityChecker.java
```diff
@@ -0,0 +1,40 @@
+package org.mskcc.cbio.portal.dao;
+
+import java.util.List;
+
+/**
+ * @deprecated Use {@link ClickHouseConstraintChecker} for ClickHouse constraint checks.
+ */
+@Deprecated
+public class ClickHouseReferentialIntegrityChecker {
```
sheridancbio, line 9: This package does contain other "non-DAO" classes, but positioning this here expands that list. First I wondered whether this class should contain a "Dao" prefix in the name ... as many other classes in this package have. But this class is not at all a Dao class ... so perhaps it belongs in another package?
Overall, I think we are not really doing DAO anymore ... and perhaps this pattern is not really appropriate for clickhouse database operations anyway. I guess I'm recommending that in the future, when we redesign the database schema, we also redesign the persistence layer (and in particular the bulk loader) to be appropriate for clickhouse operations. We should drop the package org.mskcc.cbio.portal.dao.  In fact, we should drop the package org.mskcc completely because this is not mskcc specific code anyway. But that's for the future.

src/main/java/org/mskcc/cbio/portal/scripts/CheckClickHouseConstraints.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * Checks ClickHouse foreign-key and unique-key constraints for the portal schema.
+ */
+public class CheckClickHouseConstraints extends ConsoleRunnable {
+
+    private static final int SAMPLE_VALUE_LIMIT = 5;
```
sheridancbio, line 21: Perhaps this limit should be called "EXAMPLE_VALUE_LIMIT" ... so as not to confuse the list of examples with cancer samples. Similarly all the terminology in the class referring to "sampleValues" should be instead "exampleValues"

sheridancbio (approved):
Looks good and I think this can be merged as is. Some suggestions for future work are included in comments.

Also, I suggest we move towards representing the referential integrity expectations not in code but in a specification file to live alongside the schema file.

Another recommendation (for future perhaps) is to enable a mode where *all* violations are output ... (some kind of verbose flag). That way they can all be cleaned up perhaps in one pass (although I guess cleaning up can be done with the same kind of "look for null after join" logic in SQL statements --- however, it is safer to delete explicitly referenced records rather than using general condition matching which might accidentally delete more than desired when typos/errors are made)

Note: My review of the testing section was only a cursory scan.

sheridancbio:
One followup comment/question:
The foreign key contraints find cases where a referring table refers to a non-existing record in the referred-to table. But there are also at least two tables (mutation_event and cna_event) which have records in them which are only for the purpose of being **referred to**.  If all mutation records which refer to a mutation_event record have been deleted, the mutation_event record is now effectively "dead" and should be deleted. In fact, we must remove these records for cases where we have re-annotated a mutation and want to clear out and replace the mutation_event record by deleting all studies which contain that mutation_event and then re-importing them with the new annotation.

We can do this as a follow up effort, but I think we need to detect and output records in this category.  I think we eliminate such records after each full study import or full study deletion ... but if we begin using incremental update most of the time, these records may accumulate and cause problems.

Is my concern valid @forus ?

PR #129, opened 2026-03-26
Raise an error if timeline values are duplicated

oplantalech:
Fix #128 

Raise an error if timeline entries are duplicated


oplantalech:
@sheridancbio I think this highlights a need for a broader discussion on how we handle timeline duplications. My PR targets full record duplication to keep the logic lean, but since #108 takes a case-by-case approach, we’re at a crossroads. Do we prefer a catch-all validation, or the granular control seen in #108?

PR #131, opened 2026-03-30
Fix issue with long-running deletes

jamesqo:
- Add `ClickHouseBulkDeleter` class instead of manually batching deletes
  - Uploads the IDs we want to delete into a staging ClickHouse table
  - Calls `DELETE FROM target_table WHERE id in (SELECT id FROM tmp_table)`
  - Deletes the staging table

cc @sheridancbio @forus

### DIFF
--- a/src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
@@ -1,3 +1,35 @@
+/*
+ * Copyright (c) 2026 Memorial Sloan-Kettering Cancer Center.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY, WITHOUT EVEN THE IMPLIED WARRANTY OF MERCHANTABILITY OR FITNESS
+ * FOR A PARTICULAR PURPOSE. The software and documentation provided hereunder
+ * is on an "as is" basis, and Memorial Sloan-Kettering Cancer Center has no
+ * obligations to provide maintenance, support, updates, enhancements or
+ * modifications. In no event shall Memorial Sloan-Kettering Cancer Center be
+ * liable to any party for direct, indirect, special, incidental or
+ * consequential damages, including lost profits, arising out of the use of this
+ * software and its documentation, even if Memorial Sloan-Kettering Cancer
+ * Center has been advised of the possibility of such damage.
+ */
+
+/*
+ * This file is part of cBioPortal.
+ *
+ * cBioPortal is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
 package org.mskcc.cbio.portal.dao;
 
 import java.io.ByteArrayInputStream;
--- a/src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
@@ -14,9 +14,10 @@ import java.util.List;
 import java.util.Map;
 
 /**
- * Bulk deleter that buffers IDs in memory, streams them into a ClickHouse temporary
- * table, and issues a single DELETE ... WHERE id IN (SELECT id FROM tmp) statement.
- * This avoids large numbers of parameterized IN-clause round-trips.
+ * Bulk deleter that buffers IDs in memory, streams them into a ClickHouse staging
+ * table via TSVWithNames, and issues a single DELETE ... WHERE id IN (SELECT id FROM
+ * staging_table) statement. The staging table is a real MergeTree table (not temporary)
+ * so that it persists across HTTP requests on ClickHouse Cloud.
  *
  * Mirrors the structure of ClickHouseBulkLoader.
  */
@@ -26,13 +27,13 @@ public class ClickHouseBulkDeleter {
 
     private final String targetTable;
     private final String idColumn;
-    private final String tmpTable;
+    private final String stagingTable;
     private final List<Long> pendingIds = new ArrayList<>();
 
     private ClickHouseBulkDeleter(String targetTable, String idColumn) {
         this.targetTable = targetTable;
         this.idColumn = idColumn;
-        this.tmpTable = "tmp_delete_" + targetTable;
+        this.stagingTable = "staging_delete_" + targetTable;
     }
 
     public static ClickHouseBulkDeleter getBulkDeleter(String targetTable, String idColumn) {
@@ -68,34 +69,42 @@ public class ClickHouseBulkDeleter {
         try {
             con = JdbcUtil.getDbConnection(ClickHouseBulkDeleter.class);
 
-            // 1. Create temp table
+            // Drop any leftover staging table from a previous crashed run
             try (PreparedStatement stmt = con.prepareStatement(
-                    "CREATE TEMPORARY TABLE " + tmpTable + " (id Int64) ENGINE = Memory")) {
+                    "DROP TABLE IF EXISTS " + stagingTable)) {
                 stmt.executeUpdate();
             }
 
-            // 2. Bulk-insert IDs into temp table via TSV stream
+            // Create staging table
+            try (PreparedStatement stmt = con.prepareStatement(
+                    "CREATE TABLE " + stagingTable + " (id Int64) ENGINE = MergeTree() ORDER BY id")) {
+                stmt.executeUpdate();
+            }
+
+            // Insert IDs into staging table via TSV stream
             byte[] payload = buildTsvPayload();
             try (PreparedStatement stmt = con.prepareStatement(
-                    "INSERT INTO " + tmpTable + " (id) FORMAT TSVWithNames")) {
+                    "INSERT INTO " + stagingTable + " (id) FORMAT TSVWithNames")) {
                 stmt.setBinaryStream(1, new ByteArrayInputStream(payload));
                 stmt.executeUpdate();
             }
 
-            // 3. Execute delete
+            // Execute delete via staging table
+            int deleted;
             try (PreparedStatement stmt = con.prepareStatement(
-                    "DELETE FROM " + targetTable + " WHERE " + idColumn + " IN (SELECT id FROM " + tmpTable + ")")) {
-                int deleted = stmt.executeUpdate();
-                return deleted;
+                    "DELETE FROM " + targetTable + " WHERE " + idColumn + " IN (SELECT id FROM " + stagingTable + ")")) {
+                deleted = stmt.executeUpdate();
             }
+
+            return deleted;
         } catch (SQLException | IOException e) {
             throw new DaoException(e);
         } finally {
-            // 4. Always drop temp table to avoid session state leak back into the connection pool
+            // Always drop staging table
             try {
                 if (con != null) {
                     try (PreparedStatement drop = con.prepareStatement(
-                            "DROP TEMPORARY TABLE IF EXISTS " + tmpTable)) {
+                            "DROP TABLE IF EXISTS " + stagingTable)) {
                         drop.executeUpdate();
                     }
                 }
--- a/src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
@@ -1,0 +1,121 @@
+package org.mskcc.cbio.portal.dao;
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Collection;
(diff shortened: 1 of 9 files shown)


src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private final String targetTable;
+    private final String idColumn;
+    private final String stagingTable;
+    private final List<Long> pendingIds = new ArrayList<>();
+
+    private ClickHouseBulkDeleter(String targetTable, String idColumn) {
+        this.targetTable = targetTable;
+        this.idColumn = idColumn;
+        this.stagingTable = "staging_delete_" + targetTable;
```
sheridancbio, line 68: I've posted a slack message pointing out that creating temporary / tracking / historical (like _BACKUP) tables in the main production database will lead to unnecessary and probably problematic copies of those temp tables while we are doing preparation of a new database increment. We use clone_clickhouse_database.sh as a tool to copy the tables from the production database into the standby database before importing studies and then deriving tables. We don't want to be copying these processing artifacts. The clone script already hardcodes a "skip over" of table cbioportal_sequence_state ... which the importer creates to track/set autoincrement numbers. I think we should either prohibit the addition of these "artifact" tables, or we should rewrite the clone_clickhouse_database.sh to not get a list of tables to copy from the database itself, but instead get the list from a location which defines the set of base tables for cbioportal.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public static ClickHouseBulkDeleter getBulkDeleter(String targetTable, String idColumn) {
```
sheridancbio, line 71: I wondered whether there were some tables where deletion might be desired by a composite key including more than one column. I only found what I thought were peripheral cases. I think that in our main study update processing, we can rely on either this bulk deleter or else perform the needed deletion logic in crafted sql statements.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -0,0 +1,128 @@
...
+ * Mirrors the structure of ClickHouseBulkLoader.
+ */
+public class ClickHouseBulkDeleter {
+
+    private static final Map<String, ClickHouseBulkDeleter> BULK_DELETERS = new LinkedHashMap<>();
+
+    private final String targetTable;
+    private final String idColumn;
+    private final String stagingTable;
+    private final List<Long> pendingIds = new ArrayList<>();
```
sheridancbio, line 63: This works for cases where we have an integer id number identifying the record we want to delete. There are a few cases where we might want to delete based on a set of string identifiers. One example might be bulk delete from clinical_sample ... based on a set of identified ATTR_ID's. But probably the bulk delete of records from clinical_sample are going to occur because we are deleting an entire sample (which is integer identified). So maybe this is an unlikely use case.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -0,0 +1,128 @@
...
+                    "CREATE TABLE " + stagingTable + " (id Int64) ENGINE = MergeTree() ORDER BY id")) {
+                stmt.executeUpdate();
+            }
+
+            // Insert IDs into staging table via TSV stream
+            byte[] payload = buildTsvPayload();
+            try (PreparedStatement stmt = con.prepareStatement(
+                    "INSERT INTO " + stagingTable + " (id) FORMAT TSVWithNames")) {
+                stmt.setBinaryStream(1, new ByteArrayInputStream(payload));
+                stmt.executeUpdate();
```
sheridancbio, line 121: if the Update does not successfully insert all id values, will we always get an exception thrown here?

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+
+            // Execute delete via staging table
+            int deleted;
+            try (PreparedStatement stmt = con.prepareStatement(
+                    "DELETE FROM " + targetTable + " WHERE " + idColumn + " IN (SELECT id FROM " + stagingTable + ")")) {
+                deleted = stmt.executeUpdate();
```
sheridancbio, line 128: If delete is "partial" will we get an exception? The output of a call to "executeUpdate()" is an integer indicating the number of rows affected by the update execution. What should we do if we have a list of 100 ids and we are told that "50 were deleted" ? I guess that might happen if 50 of the ids in the delete list were not present from the beginning in the table. Or it might happen if we have duplicates in the delete list. Should we be interrogating this return value?

src/test/java/org/mskcc/cbio/portal/integrationTest/dao/AbstractDaoDeleteTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Count rows in {@code table} where {@code column = value}.
+     */
+    protected long countRowsWhereEq(String table, String column, long value) throws DaoException {
+        Connection con = null;
+        try {
+            con = JdbcUtil.getDbConnection(ClickHouseBulkDeleter.class);
+            PreparedStatement ps = con.prepareStatement(
+                "SELECT count() FROM " + table + " WHERE " + column + " = ?");
```
sheridancbio, line 47: Just learned something here : clickhouse SQL syntax does not require some expression in the count() parenthesis.

src/test/java/org/mskcc/cbio/portal/integrationTest/dao/AbstractDaoDeleteTest.java
```diff
@@ -0,0 +1,77 @@
+/*
+ * This file is part of cBioPortal.
```
sheridancbio, line 2: In order to use the Affero License (or any license) the copyright owner must be identified and the license terms must be referred to. See https://www.gnu.org/licenses/agpl-3.0.en.html for the instructions on how to use the Affero license. These files are missing the copyright line.

The important words are:

> How to Apply These Terms to Your New Programs
> If you develop a new program, and you want it to be of the greatest possible use to the public, the best way to achieve this is to make it free software which everyone can redistribute and change under these terms.
> 
> To do so, attach the following notices to the program. It is safest to attach them to the start of each source file to most effectively state the exclusion of warranty; and each file should have at least the "copyright" line and a pointer to where the full notice is found.
> 
>     <one line to give the program's name and a brief idea of what it does.>
>     Copyright (C) <year>  <name of author>

src/test/java/org/mskcc/cbio/portal/integrationTest/dao/TestDaoCancerStudyDelete.java
```diff
@@ -0,0 +1,140 @@
...
+    @Before
+    public void setUp() throws DaoException {
+        DaoCancerStudy.reCacheAll();
+        study = DaoCancerStudy.getCancerStudyByStableId("study_tcga_pub");
+        assertNotNull("seed data must contain study_tcga_pub", study);
+        studyId = study.getInternalId();
+    }
+
+    @Test
+    @SuppressWarnings("deprecation")
```
sheridancbio, line 63: Maybe deprecation warnings are good ... they indicate code which is brittle and will soon break. It is better to fix our code while it is relying on deprecated code rather than failing at the point that the deprecated code is revoked.

src/test/java/org/mskcc/cbio/portal/integrationTest/dao/TestDaoCancerStudyDelete.java
```diff
@@ -0,0 +1,140 @@
...
+    public void testDeleteCancerStudy_studyIsGone() throws DaoException {
+        DaoCancerStudy.deleteCancerStudy(studyId);
+
+        assertNull("study should no longer be accessible after deletion",
+            DaoCancerStudy.getCancerStudyByStableId("study_tcga_pub"));
+    }
+
+    @Test
+    @SuppressWarnings("deprecation")
+    public void testDeleteCancerStudy_geneticProfileRowsAreGone() throws Exception {
```
sheridancbio, line 73: tests are good ... although not exhaustively complete. The goal here is to test the bulk delete functionality and it is demonstrated on a few of the primary tables affected by a cancer study deletion. That is adequate.

src/test/java/org/mskcc/cbio/portal/integrationTest/dao/TestDaoCancerStudyDelete.java
```diff
@@ -0,0 +1,140 @@
...
+public class TestDaoCancerStudyDelete extends AbstractDaoDeleteTest {
+
+    private CancerStudy study;
+    private int studyId;
+
+    @Before
+    public void setUp() throws DaoException {
+        DaoCancerStudy.reCacheAll();
+        study = DaoCancerStudy.getCancerStudyByStableId("study_tcga_pub");
+        assertNotNull("seed data must contain study_tcga_pub", study);
```
sheridancbio, line 58: The word "seed" is confusing here. The actual database seed 

https://github.com/cBioPortal/datahub/tree/master/seedDB

contains no studies an only reference data. So there must be some available test resource which gets loaded in order to set up the testing environment for these tests.

src/test/java/org/mskcc/cbio/portal/integrationTest/dao/TestDaoPatientDelete.java
```diff
@@ -0,0 +1,105 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration(locations = {"classpath:/applicationContext-dao.xml"})
+public class TestDaoPatientDelete extends AbstractDaoDeleteTest {
+
+    // Seed patient TCGA-A1-A0SB has one sample: TCGA-A1-A0SB-01
```
sheridancbio, line 43: Similar to above ... seed is ambiguous here. Aside from that, one obvious brittleness to this way of testing is that if the test database resource be modified to add additional samples for test patient TCGA-A1-A0SB then this test will break.

src/test/java/org/mskcc/cbio/portal/integrationTest/dao/TestDaoSampleDelete.java
```diff
@@ -0,0 +1,113 @@
...
+        sampleInternalId = DaoSample.getSampleByCancerStudyAndSampleId(studyId, SAMPLE_STABLE_ID).getInternalId();
+
+        Patient patient = DaoPatient.getPatientByCancerStudyAndPatientId(studyId, PATIENT_STABLE_ID);
+        assertNotNull(patient);
+        patientInternalId = patient.getInternalId();
+    }
+
+    @Test
+    public void testDeleteSamples_sampleIsGone() throws DaoException {
+        DaoSample.deleteSamples(studyId, Set.of(SAMPLE_STABLE_ID));
```
sheridancbio, line 68: Similar to the TestDaoCancerStudyDelete class .. this shows good / adequate behavior of the bulk delete functionality when deleting a sample (which also deletes referring records from other tables)

sheridancbio (approved):
This looks really good to me. One fix to make before merge is to add a copyright line (as directed by the Affero license directions) with the license notices.
In other parts of the code you can find lines like this (obviously year needs to be adjusted):

src/main/java/org/mskcc/cbio/portal/model/GenePanel.java:
> * Copyright (c) 2016 Memorial Sloan Kettering Cancer Center.

sheridancbio:
Perhaps Copyright should be assigned to the not-for-profit 503c corporation which was created as a legal entity to allow contracting and grants for the cBioPortal project (rather than Memorial Sloan Kettering Cancer Center). We can ask about this in slack.

src/test/java/org/mskcc/cbio/portal/integrationTest/dao/TestDaoCancerStudyDelete.java
jamesqo: @sheridancbio I don't think the deprecated method (removeCancerStudy) will be deleted anytime soon. But we can address this in a future PR

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
jamesqo: It's similar to something like `int deleted = sourceSet.removeAll(otherSet)`. There's no way of telling if the source table actually contains those samples or not without trying to remove them

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
jamesqo: That's correct. We are using the same logic for the bulk loader

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
jamesqo: Yeah the Deleter class is optimized for the most common use case don't think there are is a usage of deleting by attr_id just yet

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
jamesqo: Yes, we still have some deletion logic lying around that I didn't touch because they don't follow this `DELETE FROM ... WHERE id IN <set_of_ids>` format

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
jamesqo: valid, I think we can save this for a future discussion.

jamesqo:
@sheridancbio I am keeping the copyright to MSKCC for now to follow the conventions we're using in other Java files in this repo.

jamesqo:
@sheridancbio have addressed your concerns re the copyright license -- also added an extra line for clearing cache after deletes, so the unit tests don't fail. I'll go ahead and merge this now

PR #134, opened 2026-04-01
Fix failing unit tests for RC branch

jamesqo:
There was a pull request from last week that incorrectly capitalized the column names used to initialize the seed database.

/cc @sheridancbio @forus

### DIFF
--- a/src/main/java/org/mskcc/cbio/portal/dao/DaoClinicalData.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/DaoClinicalData.java
@@ -753,45 +753,4 @@ public final class DaoClinicalData {
             JdbcUtil.closeAll(DaoClinicalData.class, con, pstmt, rs);
         }
     }
-
-    private static List<Map.Entry<Integer, String>> getExistingAttributes(String tableName, List<Map.Entry<Integer, String>> internalPatientIdAttributes) throws DaoException {
-        if (internalPatientIdAttributes.isEmpty()) {
-            return Collections.emptyList();
-        }
-        Connection con = null;
-        PreparedStatement pstmt = null;
-        ResultSet rs = null;
-
-        try {
-            con = JdbcUtil.getDbConnection(DaoClinicalData.class);
-            pstmt = con.prepareStatement("SELECT internal_id, attr_id FROM `"
-                    + tableName + "` WHERE (internal_id, attr_id) IN ("
-                    + String.join(",", Collections.nCopies(internalPatientIdAttributes.size(), "(?, ?)")) + ")");
-            int parameterIndex = 1;
-            for (Map.Entry<Integer, String> entry : internalPatientIdAttributes) {
-                pstmt.setInt(parameterIndex++, entry.getKey());
-                pstmt.setString(parameterIndex++, entry.getValue());
-            }
-            rs = pstmt.executeQuery();
-
-            List<Map.Entry<Integer, String>> result = new ArrayList<>();
-            while (rs.next()) {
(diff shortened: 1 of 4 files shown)


jamesqo:
More fixes incoming. This is the latest set of failures:

```
TestIncrementalCopyNumberAlterationImport.testDiscreteCNA_long	❌ NOT FIXED	Unrelated data inconsistency issue
TestIncrementalCopyNumberAlterationImport.testDiscreteCNA_wide	❌ NOT FIXED	Data mismatch (10 vs 8 elements)
TestIncrementalProteinLevelImport.testRppa	❌ NOT FIXED	Data mismatch (10 vs 8 elements)
TestIncrementalGenericAssayImporter.testGenericAssay	❌ NOT FIXED	Data mismatch (4 vs 2 elements)
TestImportClinicalData.testImportClinicalData_CorrectFileTwice1	❌ NOT FIXED	Duplicate detection logic issue
```

and this is what Claude has cited as the cause of the latest test failures:


> All incremental genetic alteration tests: GeneticAlterationIncrementalImporter.initialize() reads old data into memory but never deletes old rows before inserting new ones. ClickHouse (unlike MySQL) has no unique key enforcement, so both old and new rows coexist. When reading back, the old genetic_profile_samples row is returned (wrong sample count), but the new genetic_alteration values have more elements → mismatch.
> 
> Clinical data patient duplicate test: The patient attribute duplicate check only runs inside addPatientToDatabase() which is only called for new patients. On second import the patient already exists, so the check is skipped and duplicates are silently inserted.

src/main/java/org/mskcc/cbio/portal/dao/DaoClinicalData.java
```diff
@@ -753,45 +753,4 @@ public static void removePatientAttributesData(int internalPatientId) throws Dao
             JdbcUtil.closeAll(DaoClinicalData.class, con, pstmt, rs);
         }
     }
-
-    private static List<Map.Entry<Integer, String>> getExistingAttributes(String tableName, List<Map.Entry<Integer, String>> internalPatientIdAttributes) throws DaoException {
```
sheridancbio, line 757: This seems to be dead code ... we might be able to clean up more if we want.

src/test/java/org/mskcc/cbio/portal/integrationTest/scripts/TestImportClinicalData.java
```diff
@@ -413,31 +413,6 @@ public void testImportClinicalData_PatientIdError() throws Exception {
...
-	 */
-	@Test
-	public void testImportClinicalData_CorrectFileTwice1() throws Exception {
-		checkCorrectFileTwice("PATIENT", "Error: The following patient internal id and attribute entries already exist in the database:");
-	}
-    @Test
-	public void testImportClinicalData_CorrectFileTwice2() throws Exception {
-		checkCorrectFileTwice("SAMPLE", "Error: Sample TEST-A2-A04P is already in the database.");
-	}	
-	private void checkCorrectFileTwice(String type, String errorMessage) throws Exception {
```
sheridancbio, line 429: I remember you mentioned that a factor here is that importing to a ReplacingMergeTree table, the table *will* accept duplicates temporarily and will eventually reach consistency by integrating records which duplicate the ORDER BY key and replacing the older record with the more recent one so that only one record exists. That must be why no exception is thrown (as under mysql).

We might want to schedule a round of unit test writing in the future which introduce tests which are appropriate for the expected behavior in the clickhouse environment.

sheridancbio (commented):
looks good 👍

PR #139, opened 2026-04-08
Adjust clickhouse clone script to handle ReplacingMergeTree tables properly

jamesqo:
If the table is backed by the ReplacingMergeTree engine, it's necessary to append the `FINAL` modifier to get an accurate count comparison -- otherwise we could be comparing merged and unmerged counts.

cc @sheridancbio


scripts/clickhouse_import_support/clone_clickhouse_database.sh
```diff
@@ -165,27 +165,50 @@ function copy_source_database_table_data_to_destination() {
...
 
-function destination_table_matches_source_table() {
-    local table_name=$1
+function get_table_record_count() {
+    local database_name=$1
+    local table_name=$2
+    # Use FINAL for ReplacingMergeTree engines (including SharedReplacingMergeTree on
+    # ClickHouse Cloud) so that deduplication is applied at read time, giving a
+    # consistent logical row count rather than the raw unmerged storage count.
+    local engine_statement="SELECT engine FROM system.tables WHERE database = '$database_name' AND name = '$table_name'"
```
sheridancbio, line 174: This works because users who are granted authorities for particular databases can access the system.tables records related to those database. Authority grants we are using for our import user are:

GRANT SHOW TABLES, SHOW COLUMNS, SHOW DICTIONARIES, SELECT, INSERT, ALTER, CREATE TABLE, CREATE VIEW, CREATE DICTIONARY, DROP TABLE, DROP VIEW, DROP DICTIONARY, TRUNCATE, OPTIMIZE, CREATE ROW POLICY, ALTER ROW POLICY, DROP ROW POLICY, SHOW ROW POLICIES, dictGet on dbname.*

scripts/clickhouse_import_support/clone_clickhouse_database.sh
```diff
@@ -165,27 +165,50 @@ function copy_source_database_table_data_to_destination() {
...
+    if ! execute_sql_statement_via_clickhouse_client "$engine_statement" "$record_count_comparison_filepath" ; then
+        echo "Warning : failed to get engine for table $database_name.$table_name" >&2
+        return 1
+    fi
+    unset sql_data_array
+    set_clickhouse_sql_data_array_from_file "$record_count_comparison_filepath" 0
+    local engine="${sql_data_array[0]}"
+    local final_clause=""
+    if [[ "$engine" == *"ReplacingMergeTree"* ]] ; then
```
sheridancbio, line 183: This is a new syntax for me ... I learned something here.
(I would have used regex.)

scripts/clickhouse_import_support/clone_clickhouse_database.sh
```diff
@@ -165,27 +165,50 @@ function copy_source_database_table_data_to_destination() {
...
+        final_clause=" FINAL"
+    fi
     # SETTINGS select_sequential_consistency = 1 ensures that count(*) sees all
     # inserted rows even on ClickHouse Cloud (SharedMergeTree / S3-backed), where
     # reads are eventually consistent and a count(*) immediately after INSERT may
     # otherwise return a stale result.
-    local statement="SELECT (SELECT count(*) FROM \`$destination_database_name\`.\`$table_name\`) = (SELECT count(*) FROM \`$source_database_name\`.\`$table_name\`), (SELECT count(*) FROM \`$source_database_name\`.\`$table_name\`), (SELECT count(*) FROM \`$destination_database_name\`.\`$table_name\`) SETTINGS select_sequential_consistency = 1"
-    if ! execute_sql_statement_via_clickhouse_client "$statement" "$record_count_comparison_filepath" ; then
-        echo "Warning : failed to execute clickhouse statement : $statement" >&2
+    local count_statement="SELECT count(*) FROM \`$database_name\`.\`$table_name\`$final_clause SETTINGS select_sequential_consistency = 1"
```
sheridancbio, line 190: Is there a penalty for using FINAL with a regular MergeTree table? If no real cost is added, we might just always use FINAL.

scripts/clickhouse_import_support/clone_clickhouse_database.sh
```diff
@@ -165,27 +165,50 @@ function copy_source_database_table_data_to_destination() {
...
+    return 0
+}
+
+function destination_table_matches_source_table() {
+    local table_name=$1
+    local src_count dst_count
+    if ! src_count=$(get_table_record_count "$source_database_name" "$table_name") ; then
+        return 1
+    fi
+    if ! dst_count=$(get_table_record_count "$destination_database_name" "$table_name") ; then
```
sheridancbio, line 207: Previously, we executed one sql statement to obtain both record counts. This approach executes (in series) two sql statements (one for source, one for destination) so execution time will be twice the time of one. But I expect that the count(*) function in clickhouse has trivial execution time (unlike in mysql with INNODB tables which could take many minutes for large tables)

scripts/clickhouse_import_support/clone_clickhouse_database.sh
```diff
@@ -165,27 +165,50 @@ function copy_source_database_table_data_to_destination() {
...
-        echo "Error : when cloning data from table $table_name, source database and destination database tables contain different record counts (source: $source_count, destination: $destination_count)" >&2
+    echo "${sql_data_array[0]}"
+    return 0
+}
+
+function destination_table_matches_source_table() {
+    local table_name=$1
+    local src_count dst_count
+    if ! src_count=$(get_table_record_count "$source_database_name" "$table_name") ; then
+        return 1
```
sheridancbio, line 205: Lets add some error reporting here to indicate why the script failed / exited with non-zero status. (like : "could not obtain a records count for source database $source_database_name.$table_name).

oh .. never mind. I see that the warning is output inside of function get_table_record_count() 👍

sheridancbio (dismissed):
LGTM 👍

scripts/clickhouse_import_support/clone_clickhouse_database.sh
jamesqo: good to know! yes I think we have other queries that access `system.tables` for metadata

scripts/clickhouse_import_support/clone_clickhouse_database.sh
jamesqo: @sheridancbio it's because clickhouse does not support using FINAL for non-replacing merge tree tables. 

Code: 181. DB::Exception: Received from mfqw5rkqsp.us-east-1.aws.clickhouse.cloud:9440. DB::Exception: Storage SharedMergeTree doesn't support FINAL. (ILLEGAL_FINAL)

PR #140, opened 2026-04-08
Remove references to MySQL from GENIE synchronize users script

jamesqo:


sheridancbio (approved):
Looks perfect

scripts/clickhouse_import_support/synchronize_user_tables_between_databases.sh
```diff
@@ -146,123 +125,92 @@ function clickhouse_destination_database_exists() {
...
 
 function find_records_missing_in_destination_database_users_table() {
-    local table_name="users"
-    local source_table_full_name="\`$mysql_source_database_name\`.\`$table_name\`"
-    local destination_table_full_name="\`$mysql_destination_database_name\`.\`$table_name\`"
-    local find_records_missing_in_destination_database_table_statement="SELECT $source_table_full_name.EMAIL, $source_table_full_name.NAME, $source_table_full_name.ENABLED FROM $source_table_full_name LEFT JOIN $destination_table_full_name ON $source_table_full_name.EMAIL=$destination_table_full_name.EMAIL WHERE $destination_table_full_name.EMAIL IS NULL;"
-    if ! execute_sql_statement_via_mysql "$find_records_missing_in_destination_database_table_statement" "$missing_user_records_result_filepath" ; then
+    local source_table="\`$clickhouse_source_database_name\`.\`users\`"
+    local dest_table="\`$clickhouse_destination_database_name\`.\`users\`"
+    local statement="SELECT src.email, src.name, src.enabled FROM $source_table AS src LEFT JOIN $dest_table AS dest ON src.email = dest.email WHERE dest.email = '';"
```
sheridancbio, line 130: This used to be "IS NULL" in mysql dialect ... I think the equivalent clickhouse function is "isNull(dest.email)", but perhaps that is equivalent to:
dest.email = ''. I guess we could test whether an empty string can be distinguished from a null value in a string field.

PR #144, opened 2026-04-21
Disable backup functionality for non-incremental update mode

jamesqo:
see https://cbioportal.slack.com/archives/C09GX3CMKFU/p1776788511579739

### DIFF
--- a/src/main/java/org/mskcc/cbio/portal/dao/BackupUtil.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/BackupUtil.java
@@ -15,7 +15,22 @@ public class BackupUtil {
         void run() throws Exception;
     }
 
-    public static void withBackup(List<String> tableNames, ThrowingRunnable fn) throws Exception {
+    public static void conditionalBackup(
+        boolean isIncrementalUpdateMode,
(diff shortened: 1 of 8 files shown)


src/main/java/org/mskcc/cbio/portal/dao/BackupUtil.java
```diff
@@ -15,7 +15,22 @@ public interface ThrowingRunnable {
...
+        List<String> tableNames,
+        ThrowingRunnable fn
+    ) throws Exception {
+
+        // We don't need to worry about backing up tables
+        // if we're not running in incremental update mode--
+        // the new data will be uploaded in a single SQL query,
+        // and even if there is an interruption mid-import the data
+        // should not be left in an inconsistent state.
+        if (!isIncrementalUpdateMode) {
```
sheridancbio, line 29: @jamesqo the other case where backup is needed is when the user has called into cbioportalImporter.py with a command such as:
- "remove-samples"
- "remove-patients"

During the processing of these removals, a similar "partially modified" problem could occur and it would be well advised for us to use the backup/restore functionality there too.

PR #146, opened 2026-05-01
Fix Missing Mutation Count

averyniceday:
- delete is in race condition deleting newly inserted clinical records
- rely on mergetree to deduplicate and keep the latest records

### DIFF
--- a/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
@@ -45,6 +45,8 @@ import org.mskcc.cbio.portal.model.shared.GeneticAlterationType;
 import org.mskcc.cbio.portal.model.GeneticProfile;
 import org.mskcc.cbio.portal.model.Sample;
 import org.mskcc.cbio.portal.util.MutationKeywordUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.sql.Connection;
 import java.sql.PreparedStatement;
@@ -66,6 +68,7 @@ import java.util.regex.Pattern;
  * Data access object for Mutation table
  */
 public final class DaoMutation {
+    private static final Logger LOG = LoggerFactory.getLogger(DaoMutation.class);
     public static final String NAN = "NaN";
     private static final String MUTATION_COUNT_ATTR_ID = "MUTATION_COUNT";
     private static final String DELETE_ALTERATION_DRIVER_ANNOTATION = "DELETE from alteration_driver_annotation WHERE genetic_profile_id=? and sample_id=?";
@@ -170,8 +173,7 @@ public final class DaoMutation {
         }
 
         Connection con = null;
-        PreparedStatement deleteStmt = null;
-        PreparedStatement insertStmt = null;
+        PreparedStatement selectStmt = null;
         ResultSet rs = null;
         try {
             con = JdbcUtil.getDbConnection(DaoMutation.class);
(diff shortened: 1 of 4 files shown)


src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -183,8 +185,14 @@ public static void createMutationCountClinicalData(GeneticProfile geneticProfile
                 DaoClinicalAttributeMeta.addDatum(attr);
             }
 
-            String mutationCountSelect =
-                "SELECT sample_profile.`sample_id`, 'MUTATION_COUNT', count(DISTINCT mutation_event.`chr`, mutation_event.`start_position`, " +
+            LOG.info("createMutationCountClinicalData (bulk insert) for profile {}", geneticProfile.getGeneticProfileId());
```
sheridancbio, line 188: The genetic_profile_id is the internal ID ... which will be reset the next time the profile is imported to. It would be more informative to log the profile stable_id.

src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -193,29 +201,26 @@ public static void createMutationCountClinicalData(GeneticProfile geneticProfile
...
-            );
-            insertStmt.setInt(1, geneticProfile.getGeneticProfileId());
-            insertStmt.executeUpdate();
+            LOG.info("inserting MUTATION_COUNT for {} samples", mutationCounts.size());
+            for (Map.Entry<Integer, String> entry : mutationCounts.entrySet()) {
+                DaoClinicalData.addSampleDatum(entry.getKey(), MUTATION_COUNT_ATTR_ID, entry.getValue());
+            }
 
-            // Necessary for deduplication
+            // Flush inserts and force ReplacingMergeTree deduplication so the latest values are visible
```
sheridancbio, line 218: I was not very familiar with this ... I guess calling optimizeTables() involves (inherently) a flushing of all data added to the bulk loader?. I guess that is good. One thing I am concerned about is the frequent call to OPTIMIZE TABLE <tablename> FINAL. For very large tables (such as genetic_alterations) this operation requires a significant amount of memory and may fail for reasons of insufficient memory. In the future, I would like to explore using other ways to avoid this or maybe simply not calling the OPTIMIZE request if it is not needed --- so perhaps first checking for duplicated records in the data being changed in the current run.

sheridancbio (approved):
I think this fix makes sense. We should survey our other code to see whether this same kind of "eventual consistency of an asynchronous data change request" is present and we have other race condition vulnerabilities.

Another concern I have (and which should be tested) is the removal of the deletion operation. When we are populating a new study (a full study import) there should be no clinical attribute called "MUTATION_COUNT" there which needs to be deleted --- but hypothetically there could be if there is a column in the data_clinical_sample.txt file which has values under such a column. We could test that. I wonder if then we would wind up with two records per sample for this ATTR_ID (MUTATION_COUNT)? Maybe not if the clinical_sample table is ReplacingMergeTree ... but we should test. But more importantly, we should test this for an incremental study update. If a sample in the study had (let's say) 5 mutation events before ... but was updated during an incremental update operation to load in the new mutations (which has 7 mutation events) .. would we then wind up with two records for the sample defining the MUTATION_COUNT as both 5 and 7 ? Or maybe it would fail with an error.

jamesqo:
I'm going to merge this so I can rebuild the jarfiles. Note: This PR assumes that there will always be replacement MUTATION_COUNT data for all the samples in question, it will break if there isn't replacement data

PR #149, opened 2026-05-07
drop async settings for derived tables

sheridancbio:
- also pause 90 seconds after each OPTIMIZE TABLE x FINAL statement (too many rapid calls to OPTIMIZE TABLE causes out of memory failure)
- in order to get the CI tests to pass, all scripts which formed a connection to a mysql database needed to be crippled/commented out. These scripts are no longer functional and will need to be adapted to clickhouse in order to make them functional again:
    - updateOncokbAnnotations.py
    - migrate_db.py

### DIFF
--- a/scripts/clickhouse_import_support/create_derived_tables_in_clickhouse_database.sh
+++ b/scripts/clickhouse_import_support/create_derived_tables_in_clickhouse_database.sh
@@ -235,20 +235,6 @@ function sql_statement_includes_insert_into() {
     return 1
 }
 
-function append_async_settings_suffix() {
-    local sql_filepath="$1"
-    local sql_filepath_amended="$sql_filepath.amended"
-    local async_settings_suffix=" SETTINGS async_insert=1, wait_for_async_insert=1, async_insert_busy_timeout_ms=300000, async_insert_max_data_size=$clickhouse_max_memory_use_target, async_insert_max_query_number=10000;"
-    local linecount=$(cat $sql_filepath | wc -l)
-    local linecount_minus_one=$(($linecount-1))
-    head -n $linecount_minus_one $sql_filepath > $sql_filepath_amended
-    local last_line_from_file="$(tail -n 1 $sql_filepath)"
-    local line_without_semicolon="${last_line_from_file%;*}"
-    echo "$line_without_semicolon" >> $sql_filepath_amended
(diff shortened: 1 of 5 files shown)


scripts/clickhouse_import_support/create_derived_tables_in_clickhouse_database.sh
```diff
@@ -269,15 +255,17 @@ function create_all_derived_tables() {
...
-        if sql_statement_includes_insert_into $sql_filepath ; then
-            append_async_settings_suffix $sql_filepath
-        fi
         if ! execute_sql_statement_from_file_via_clickhouse_client "$sql_filepath" "create_derived_table_result_filepath" ; then
             echo "Error : failure occurred during execution of sql statements in file $sql_filepath" >&2
             echo "    beginning with:" >&2
             head -n 4 $sql_filepath >&2
             return 1
         fi
+        if grep OPTIMIZE "$sql_filepath" | grep FINAL ; then
```
jamesqo, line 264: thought we were not changing this script? or is this just cleanup?

I think we want to replace it with `--multiquery` at some point

jamesqo (approved):

scripts/clickhouse_import_support/create_derived_tables_in_clickhouse_database.sh
sheridancbio: --multiquery will process all of the SQL statements as they appear in the source file sequentially. Doing that would fail. Our production database contains genetic profiles which contain non-matching comma separated lists for a handful of studies which are "non-rectangular". These studies have cases such as having a genetic_profile_samples list with 1000 items in the list, which would be paired up against a number of genetic_alteration event lists with 600 items in those lists (or some other number). Cases like this must be detected and filtered out of the operation in order for the derived table construction process to reach completion.

Also, there are some statements which require special treatment --- such as these "OPTIMIZE TABLE genetic_alteration FINAL" style statements. If we do 5 sequential executions like that, we observe failures because of memory depletion. So these lines introduce pauses between each statement. Rather than a fixed pause we could instead do repeated queries waiting for the OPTIMIZE TABLE statement to be fully processed (perhaps querying the list of running jobs on the server). In any case, for the time being we must do some special processing at various points in the clickhouse.sql script.

PR #150, opened 2026-05-08
Backup ga tables before sample removal

forus:
The genetic alteration tables, genetic_alteration and genetic_profile_samples, are not crash-safe.
If sample removal fails partway through, the tables can be left in an inconsistent state,
with some rows updated and others still referencing the removed sample.
    
Reuse the backup-and-restore approach already used by incremental upload: back up the affected tables
before removal and restore them if an error occurs.
    
This code path is used when removing a sample and when removing a patient.

### DIFF
--- a/src/main/java/org/mskcc/cbio/portal/dao/BackupUtil.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/BackupUtil.java
@@ -4,8 +4,6 @@ import org.mskcc.cbio.portal.util.ProgressMonitor;
 
 import java.sql.Connection;
 import java.sql.SQLException;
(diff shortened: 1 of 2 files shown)


sheridancbio (approved):
This looks good to me. There is also the possibility of deleting a patient ... and when deleting a patient, I assume we will be deleting the samples for that patient. I am guessing that deleting the samples for the deleted patient is done in a way which will also invoke removeSamplesInGeneticAlterationsForStudy() --- but we should make sure that is true.

One other thought --- generic assay data is stored in a similar way --- maybe in the same genetic_alteration table. I wonder if that is all covered by this code as well.

PR #161, opened 2026-06-08
Add retry logic to clone_clickhouse_database in order to fix ZooKeeper race condition

jamesqo:
On ClickHouse Cloud (SharedMergeTree), dropping a table is asynchronous in ZooKeeper. If a table is recreated and data is inserted immediately after, the new part metadata can collide with stale paths left over from the dropped table, causing the INSERT to fail.

This fix adds retry logic to copy_source_database_table_data_to_destination: the INSERT is retried up to 3 times with a 30-second delay between attempts, giving ZooKeeper enough time to finish cleanup before the next attempt.

cc @sheridancbio @averyniceday

### DIFF
--- a/scripts/clickhouse_import_support/clone_clickhouse_database.sh
+++ b/scripts/clickhouse_import_support/clone_clickhouse_database.sh
@@ -158,11 +158,28 @@ function create_destination_database_table_schema_only() {
 
 function copy_source_database_table_data_to_destination() {
     local table_name=$1
+    # Retry the INSERT up to 3 times with a 30-second delay between attempts.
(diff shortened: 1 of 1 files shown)


sheridancbio (approved):
The delay logic seems fine. I do not think this error has anything to do with zookeeper aside from zookeeper being software which helps coordinate the multiple running hosts which are working in coordination with each other. The real problem is that changes in clickhouse are performed locally on one node, but those changes need to propagate to other nodes (which become eventually consistent with the change). I think it is possible to set up a multi-node cluster without zookeeper, and the same kind of problems would occur in that situation.

PR #166, opened 2026-07-02
fix delete race condition bug

sheridancbio:
- rather than processing per table, process per phase (create/populate/delete/drop)
- introduce wait loops to verify that all replicas (in clusters) "see" the data in staging tables

## Problem description

In the production databases (running in a multi-replica cluster deployed through clickhouse.cloud), we noticed accumulations of "orphaned" records where a study was deleted, but associated linked records in other tables (such as the samples table) failed to be deleted. This leads to (for example) sample records which refer to a non-existent patient record in field `patient_id`.

After analysis, a race condition was suspected based on this theory of behavior in ClickHouseBulkDeleter:
- the list of internal identifiers which should be deleted are uploaded into a temporary holding table
- a sql statement which retrieves the list of internal identifiers from the table is then immediately run to delete the referenced records
- hypothesis : the first operation takes time to complete processing and for the metadata associated with the holding table to be updated to refer the the data parts which are in the process of being created ... during that processing time, the second operation begins and says "delete all records with an internal identifier present in the holding table. This operation "sees" an empty holding table and deletes nothing. This is more likely to occur when the holding table has a great number of internal identifiers being inserted. It may also be due to a multi-replica Clickhouse service (like in Clickhouse.cloud) needing to propagate the metadata updates to other processing nodes (where the second operation may have already started)

## Solution

- give more time for the holding tables to be completed by:
    - writing all holding table contents first (when multiple tables are being deleted from) before using any of the holding table contents. By doing this, by the time the final holding table write operation is started, the first table (written a while ago) will have had time to be finalized and content registered.
    - testing each holding table for proper recordcount across all processing nodes (clickhouse has a "run command on all nodes and report all results" function) and entering a short pause/retry loop until the proper counts are seen

### DIFF
--- a/src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
@@ -70,11 +70,11 @@ public class ClickHouseBulkDeleter {
     private ClickHouseBulkDeleter(String targetTable, String idColumn) {
         this.targetTable = targetTable;
         this.idColumn = idColumn;
-        this.stagingTable = "staging_delete_" + targetTable;
+        this.stagingTable = String.format("staging_delete_%s", targetTable);
     }
 
     public static ClickHouseBulkDeleter getBulkDeleter(String targetTable, String idColumn) {
-        String key = targetTable + ":" + idColumn;
+        String key = String.format("%s:%s", targetTable, idColumn);
         return BULK_DELETERS.computeIfAbsent(key, k -> new ClickHouseBulkDeleter(targetTable, idColumn));
     }
 
@@ -88,8 +88,8 @@ public class ClickHouseBulkDeleter {
         }
     }
 
-    public static int flushAll() throws DaoException {
-        int totalDeleted = 0;
+    public static long flushAll() throws DaoException {
+        long totalDeleted = 0;
         try {
             dropAnyExistingStagingTables(false); // drop any leftover tables from previous crash/failure
             createAllStagingTables();
@@ -101,7 +101,8 @@ public class ClickHouseBulkDeleter {
             try {
                 dropAnyExistingStagingTables(true);
             } catch (SQLException se) {
-                // exceptions will be ignored during second call to dropAnyExistingStagingTables
+                // will not happen because SQLExceptions will be suppressed during second call
+                // to dropAnyExistingStagingTables (leftover tables will be tolerated)
             }
             clearAllDeleters();
         }
@@ -152,7 +153,10 @@ public class ClickHouseBulkDeleter {
         //     running in a Clickhouse cluster (such as with clickhouse.cloud). This may avoid retry cycles.
         for (ClickHouseBulkDeleter deleter : BULK_DELETERS.values()) {
             if (! deleter.allReplicasReachedRecordCount()) {
-                throw new DaoException("Failed to see all replicas reflect delete list inserted into table " + deleter.stagingTable);
+                String exceptionMsgString = String.format(
+                        "Failed to see all replicas reflect delete list inserted into table %s",
+                        deleter.stagingTable);
+                throw new DaoException(exceptionMsgString);
             }
         }
     }
@@ -180,7 +184,6 @@ public class ClickHouseBulkDeleter {
         int WAIT_CYCLE_PERIOD_SECONDS = 10;
         int WAIT_CYCLE_TOLERATE_EXCEPTION_LIMIT = 6;
         int exceptions_ignored = 0;
-        log.warn(String.format("beginning on wait for table %s", stagingTable));
         for (int cycle = 0 ; cycle < WAIT_CYCLE_MAX_COUNT ; cycle = cycle + 1) {
             String getRecordCountsString = String.format(
                     "SELECT hostname() as host, COUNT() as record_count FROM clusterAllReplicas('default', current_database(), %s) GROUP BY host",
@@ -190,9 +193,13 @@ public class ClickHouseBulkDeleter {
                     ResultSet rs = pstmt.executeQuery()) {
                 boolean recordCountWasReached = true;
                 while (rs.next()) {
-                    int recordCountOnReplica = rs.getInt("record_count");
-                    log.warn(String.format("waiting to reach record count : expected %d saw %d", recordCountOnReplica, pendingIds.size()));
+                    long recordCountOnReplica = rs.getInt("record_count");
                     if (recordCountOnReplica != pendingIds.size()) {
+                        log.warn(String.format(
+                                "waiting to reach record count in table %s : expected %d saw %d",
+                                stagingTable,
+                                pendingIds.size(),
+                                recordCountOnReplica));
                         recordCountWasReached = false;
                         break;
                     }
@@ -215,16 +222,16 @@ public class ClickHouseBulkDeleter {
         return false;
     }
 
-    private static int deleteRecordsReferencedInStagingTables() throws SQLException {
-        int totalDeleted = 0;
+    private static long deleteRecordsReferencedInStagingTables() throws SQLException {
+        long totalDeleted = 0;
         for (ClickHouseBulkDeleter deleter : BULK_DELETERS.values()) {
             totalDeleted += deleter.deleteRecordsReferencedInStagingTable();
         }
         return totalDeleted;
     }
 
-    private int deleteRecordsReferencedInStagingTable() throws SQLException {
-        int records_deleted;
+    private long deleteRecordsReferencedInStagingTable() throws SQLException {
+        long records_deleted;
         String statementString = String.format(
                 "DELETE FROM %s WHERE %s IN (SELECT id FROM %s)",
                 targetTable, idColumn, stagingTable);
@@ -237,16 +244,65 @@ public class ClickHouseBulkDeleter {
         return records_deleted;
     }
 
-    private static void dropAnyExistingStagingTables(boolean ignoreExceptions) throws SQLException {
+    private boolean allReplicasCompletedDeletion() throws SQLException {
+        int WAIT_CYCLE_MAX_COUNT = 60;
+        int WAIT_CYCLE_PERIOD_SECONDS = 10;
+        int WAIT_CYCLE_TOLERATE_EXCEPTION_LIMIT = 6;
+        int exceptions_ignored = 0;
+        for (int cycle = 0 ; cycle < WAIT_CYCLE_MAX_COUNT ; cycle = cycle + 1) {
+            String getUndeletedRecordCountsString = String.format(
+                    "SELECT hostname() as host, COUNT() as record_count FROM clusterAllReplicas('default', current_database(), %s) WHERE %s IN (SELECT id from %s) GROUP BY host",
+                    targetTable, idColumn, stagingTable);
+            Connection con = JdbcUtil.getDbConnection(ClickHouseBulkDeleter.class);
+            try (PreparedStatement pstmt = con.prepareStatement(getUndeletedRecordCountsString);
+                    ResultSet rs = pstmt.executeQuery()) {
+                boolean allRecordsWereDeleted = true;
+                while (rs.next()) {
+                    long undeletedRecordCountOnReplica = rs.getInt("record_count");
+                    if (undeletedRecordCountOnReplica != 0) {
+                        allRecordsWereDeleted = false;
+                        break;
+                    }
+                }
+                if (allRecordsWereDeleted) {
+                    return true;
+                }
+                Thread.sleep(1000 * WAIT_CYCLE_PERIOD_SECONDS);
+            } catch (SQLException e) {
+                if (exceptions_ignored >= WAIT_CYCLE_TOLERATE_EXCEPTION_LIMIT) {
+                    log.error("Over the limit of exceptions in allreplicasCompletedDeletion() .. re-throwing " + e.getMessage());
+                    throw e;
+                }
+                exceptions_ignored = exceptions_ignored + 1;
+            } catch (InterruptedException ie) {
+                // ignore : ok to go on to next cycle immediately
+            } finally {
+                JdbcUtil.closeAll(ClickHouseBulkDeleter.class, con, null, null);
+            }
+        }
+        return false;
+    }
+    private static void dropAnyExistingStagingTables(boolean deletionHasBeenExecuted) throws SQLException, DaoException {
         for (ClickHouseBulkDeleter deleter : BULK_DELETERS.values()) {
+            // wait until all ids in the stagingTable table have been fully deleted from the target table
+            if (deletionHasBeenExecuted && ! deleter.allReplicasCompletedDeletion()) {
+                String exceptionMessageString = String.format(
+                        "Failed to complete the delete operation on all replicas for table %s",
+                        deleter.stagingTable);
+                throw new DaoException(exceptionMessageString);
+            }
             String dropStatementString = String.format(
-                "DROP TABLE IF EXISTS %s",
-                deleter.stagingTable);
+                    "DROP TABLE IF EXISTS %s",
+                    deleter.stagingTable);
             Connection con = JdbcUtil.getDbConnection(ClickHouseBulkDeleter.class);
             try (PreparedStatement stmt = con.prepareStatement(dropStatementString)) {
                 stmt.executeUpdate();
             } catch (SQLException e) {
-                if (! ignoreExceptions) {
+                if (! deletionHasBeenExecuted) {
+                    // it is fatal to fail to DROP any old staging tables during the first call
+                    // (during setup for deletion), because then the CREATE for staging will fail.
+                    // on the second call, so long as the deletion completed, it is "ok" to leave behind
+                    // residual staging tables -- they will (hopefully) DROP during the next update run
                     throw e;
                 }
             } finally {
--- a/src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
@@ -45,6 +45,8 @@ import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Bulk deleter that buffers IDs in memory, streams them into a ClickHouse staging
@@ -62,6 +64,8 @@ public class ClickHouseBulkDeleter {
     private final String idColumn;
     private final String stagingTable;
     private final Set<Long> pendingIds = new HashSet<>();
+
+    private static final Logger log = LoggerFactory.getLogger(ClickHouseBulkDeleter.class);
 
     private ClickHouseBulkDeleter(String targetTable, String idColumn) {
         this.targetTable = targetTable;
@@ -86,21 +90,16 @@ public class ClickHouseBulkDeleter {
 
     public static int flushAll() throws DaoException {
         int totalDeleted = 0;
-        Connection con = null;
         try {
-            con = JdbcUtil.getDbConnection(ClickHouseBulkDeleter.class);
-            dropAnyExistingStagingTables(con, false); // drop any leftover tables from previous crash/failure
-            createAllStagingTables(con);
-            populateAllStagingTables(con);
-            totalDeleted = deleteRecordsReferencedInStagingTables(con);
+            dropAnyExistingStagingTables(false); // drop any leftover tables from previous crash/failure
+            createAllStagingTables();
+            populateAllStagingTables();
+            totalDeleted = deleteRecordsReferencedInStagingTables();
         } catch (SQLException | IOException e) {
             throw new DaoException(e);
         } finally {
             try {
-                if (con != null) {
-                    dropAnyExistingStagingTables(con, true);
-                    JdbcUtil.closeAll(ClickHouseBulkDeleter.class, con, null, null);
-                }
+                dropAnyExistingStagingTables(true);
             } catch (SQLException se) {
                 // exceptions will be ignored during second call to dropAnyExistingStagingTables
             }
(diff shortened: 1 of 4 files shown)


src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -84,66 +85,142 @@ public void addIds(Collection<? extends Number> ids) {
...
+        for (ClickHouseBulkDeleter deleter : BULK_DELETERS.values()) {
+            totalDeleted += deleter.deleteRecordsReferencedInStagingTable(con);
+        }
+        return totalDeleted;
+    }
+
+    private int deleteRecordsReferencedInStagingTable(Connection con) throws SQLException {
+        int records_deleted;
+        try (PreparedStatement stmt = con.prepareStatement(
+                "DELETE FROM ? WHERE ? IN (SELECT id FROM ?)")) {
```
jamesqo: @sheridancbio Has this part been tested? I asked my agent and it flagged a potential issue here:

```
DELETE FROM ? WHERE ? IN (SELECT id FROM ?) — parameterized identifiers don't work in JDBC

Line 204-208:java
PreparedStatement stmt = con.prepareStatement(
    "DELETE FROM ? WHERE ? IN (SELECT id FROM ?)");
stmt.setString(1, targetTable);
stmt.setString(2, idColumn);
stmt.setString(3, stagingTable);




JDBC PreparedStatement ? placeholders are for data values only — you cannot parameterize table names, column names, or other SQL identifiers this way. The JDBC driver will quote the strings as literals, producing broken SQL like:sql
DELETE FROM 'samples' WHERE 'patient_id' IN (SELECT id FROM 'staging_delete_samples')




This is a syntax error. Every other DELETE in the codebase (e.g. DaoStructuralVariant.java:163) uses string concatenation for table/column identifiers and ? only for values. The old code was correct:java
"DELETE FROM " + targetTable + " WHERE " + idColumn + " IN (SELECT id FROM " + stagingTable + ")"
```


Fix: Revert to string concatenation for identifiers.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -84,66 +85,142 @@ public void addIds(Collection<? extends Number> ids) {
...
             }
-
             // Insert IDs into staging table via TSV stream
             byte[] payload = buildTsvPayload();
-            try (PreparedStatement stmt = con.prepareStatement(
-                    "INSERT INTO " + stagingTable + " (id) FORMAT TSVWithNames")) {
-                stmt.setBinaryStream(1, new ByteArrayInputStream(payload));
-                stmt.executeUpdate();
-            }
+            PreparedStatement stmt = con.prepareStatement(
```
jamesqo: use try-with-resources

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -84,66 +85,142 @@ public void addIds(Collection<? extends Number> ids) {
...
         }
+        BULK_DELETERS.clear();
+    }
 
-        Connection con = null;
-        try {
-            con = JdbcUtil.getDbConnection(ClickHouseBulkDeleter.class);
+    private static void createAllStagingTables(Connection con) throws SQLException {
+        for (ClickHouseBulkDeleter deleter : BULK_DELETERS.values()) {
+            PreparedStatement stmt = con.prepareStatement(
```
jamesqo: use try-with-resources

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -84,66 +85,142 @@ public void addIds(Collection<? extends Number> ids) {
...
+                    "CREATE TABLE " + deleter.stagingTable + " (id Int64) ENGINE = MergeTree() ORDER BY id");
+            stmt.executeUpdate();
+        }
+        // wait for table existence to propagate to replicas (if they exist)
+        for (ClickHouseBulkDeleter deleter : BULK_DELETERS.values()) {
+            PreparedStatement stmt = con.prepareStatement(
```
jamesqo: use try-with-resources

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -84,66 +85,142 @@ public void addIds(Collection<? extends Number> ids) {
...
-                    "DELETE FROM " + targetTable + " WHERE " + idColumn + " IN (SELECT id FROM " + stagingTable + ")")) {
-                deleted = stmt.executeUpdate();
+    private boolean allReplicasReachedRecordCount(Connection con) throws SQLException {
+        int WAIT_CYCLE_MAX_COUNT = 60;
+        int WAIT_CYCLE_PERIOD_SECONDS = 10;
+        int WAIT_CYCLE_TOLERATE_EXCEPTION_LIMIT = 6;
+        int exceptions_ignored = 0;
+        for (int cycle = 0 ; cycle < WAIT_CYCLE_MAX_COUNT ; cycle = cycle + 1) {
+            try {
+                PreparedStatement pstmt = con.prepareStatement(
```
jamesqo: use try-with-resources

jamesqo (changes requested):

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: good catch. I've fixed this. I will be testing today

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -83,78 +88,238 @@ public void addIds(Collection<? extends Number> ids) {
...
+            try (PreparedStatement pstmt = con.prepareStatement(getRecordCountsString);
+                    ResultSet rs = pstmt.executeQuery()) {
+                boolean recordCountWasReached = true;
+                while (rs.next()) {
+                    long recordCountOnReplica = rs.getInt("record_count");
+                    if (recordCountOnReplica != pendingIds.size()) {
+                        log.warn(String.format(
+                                "waiting to reach record count in table %s : expected %d saw %d",
+                                stagingTable,
+                                recordCountOnReplica,
```
jamesqo: The arguments are swapped here -- this one should be last

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -83,78 +88,238 @@ public void addIds(Collection<? extends Number> ids) {
...
-                if (con != null) {
-                    try (PreparedStatement drop = con.prepareStatement(
-                            "DROP TABLE IF EXISTS " + stagingTable)) {
-                        drop.executeUpdate();
+            JdbcUtil.closeAll(ClickHouseBulkDeleter.class, con, null, null);
+        }
+        return records_deleted;
+    }
+
+    private boolean allReplicasCompletedDeletion() throws SQLException {
```
jamesqo: Maybe it's possible to refactor and consolidate code with the `allReplicasReachedRecordCount` function?

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -83,78 +88,238 @@ public void addIds(Collection<? extends Number> ids) {
...
+        for (int cycle = 0 ; cycle < WAIT_CYCLE_MAX_COUNT ; cycle = cycle + 1) {
+            String getUndeletedRecordCountsString = String.format(
+                    "SELECT hostname() as host, COUNT() as record_count FROM clusterAllReplicas('default', current_database(), %s) WHERE %s IN (SELECT id from %s) GROUP BY host",
+                    targetTable, idColumn, stagingTable);
+            Connection con = JdbcUtil.getDbConnection(ClickHouseBulkDeleter.class);
+            try (PreparedStatement pstmt = con.prepareStatement(getUndeletedRecordCountsString);
+                    ResultSet rs = pstmt.executeQuery()) {
+                boolean allRecordsWereDeleted = true;
+                while (rs.next()) {
+                    long undeletedRecordCountOnReplica = rs.getInt("record_count");
```
jamesqo: should be a getLong (other places too)

jamesqo (approved):

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -83,78 +88,238 @@ public void addIds(Collection<? extends Number> ids) {
...
+    }
+
+    private boolean allReplicasReachedRecordCount() throws SQLException {
+        int WAIT_CYCLE_MAX_COUNT = 60;
+        int WAIT_CYCLE_PERIOD_SECONDS = 10;
+        int WAIT_CYCLE_TOLERATE_EXCEPTION_LIMIT = 6;
+        int exceptions_ignored = 0;
+        for (int cycle = 0 ; cycle < WAIT_CYCLE_MAX_COUNT ; cycle = cycle + 1) {
+            String getRecordCountsString = String.format(
+                    "SELECT hostname() as host, COUNT() as record_count FROM clusterAllReplicas('default', current_database(), %s) GROUP BY host",
```
jamesqo: Claude flagged this initially saying that it might not work outside of ClickHouse Cloud, then later backtracked and said it might work OK locally. Just to be safe -- I'd do a quick test with a local Docker Compose setup, if you haven't already

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: Yes, I had the same thought. I will do it.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: good catch 👍

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: right .. will fix this

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: Yes, I have run the test (local with a docker deployed single-node clickhouse service, using the same version of clickhouse as our docker compose configuration (clickhouse/clickhouse-server:24.10)). It ran smoothly and no orphan records were remaining after study deletion. This method did not generate an error. Having said that, I agree that it is a little "heavyhanded" to be doing this cluster-aware query in order to make sure that all nodes/executors have had their metadata updated to reflect the mutated table content. An alternative (which might be preferable) was used in the PR @dippindots put together for the database migration script implementation. That monitors the system database table "mutations" looking for any mutations which are still in-progress and waiting until there are none. 
[see-code-here](https://github.com/cBioPortal/cbioportal/pull/12272/changes#diff-070da55add4b68f4ccba57b7728de65fe811c59b564ab646ef3addffc11805b4R217)
I may look into the full semantics of this approach. I wonder whether "is_done" is set to "done" at the time that the data parts are written to disk, or whether it also means that the metadata for the table has been updated on all hosts. It seems to me that it may be possible for the disk writing to be done but the metadata not yet become updated on all nodes/hosts. In that case, some nodes may still see the old state and we still may be exposed to the race condition.  I'll try to look into this.

src/main/java/org/mskcc/cbio/portal/dao/DaoPatient.java
```diff
@@ -253,7 +253,7 @@ public static void deletePatients(int internalStudyId, Set<String> patientStable
         Connection con = null;
         try {
             con = JdbcUtil.getDbConnection(DaoPatient.class);
-            List<Integer> clinicalEventIds = collectIds(con,
+            List<Long> clinicalEventIds = collectIds(con,
```
forus, line 256: Looks good.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -83,78 +88,238 @@ public void addIds(Collection<? extends Number> ids) {
...
     }
 
     private byte[] buildTsvPayload() throws IOException {
-        ByteArrayOutputStream buffer = new ByteArrayOutputStream();
-        // header row
-        buffer.write("id\n".getBytes(StandardCharsets.UTF_8));
-        // data rows
-        for (Long id : pendingIds) {
-            buffer.write((id.toString() + "\n").getBytes(StandardCharsets.UTF_8));
+        try (ByteArrayOutputStream buffer = new ByteArrayOutputStream()) {
```
forus, line 628: making sure we close/release the buffer. Looks good.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -83,78 +88,238 @@ public void addIds(Collection<? extends Number> ids) {
...
-        Connection con = null;
-        try {
-            con = JdbcUtil.getDbConnection(ClickHouseBulkDeleter.class);
-
-            // Drop any leftover staging table from a previous crashed run
-            try (PreparedStatement stmt = con.prepareStatement(
-                    "DROP TABLE IF EXISTS " + stagingTable)) {
+    private static void createAllStagingTables() throws SQLException {
+        for (ClickHouseBulkDeleter deleter : BULK_DELETERS.values()) {
+            Connection con = JdbcUtil.getDbConnection(ClickHouseBulkDeleter.class);
```
forus: It’s unusual to see the connection opened and closed inside the loop. Why not manage it outside the loop instead? Is this necessary to ensure everything is flushed properly?

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -83,78 +88,238 @@ public void addIds(Collection<? extends Number> ids) {
...
-
-            // Create staging table
-            try (PreparedStatement stmt = con.prepareStatement(
-                    "CREATE TABLE " + stagingTable + " (id Int64) ENGINE = MergeTree() ORDER BY id")) {
+        }
+        // wait for table existence to propagate to replicas (if they exist)
+        for (ClickHouseBulkDeleter deleter : BULK_DELETERS.values()) {
+            Connection con = JdbcUtil.getDbConnection(ClickHouseBulkDeleter.class);
+            String alterTableStatementString = String.format(
+                    "ALTER TABLE %s MODIFY COMMENT 'Transient' SETTINGS mutations_sync = 3",
```
forus: Why not set it for all subsequent queries in this session instead?
`SET mutations_sync = 3;`

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
forus: codex flaged this one as well: Empty replicas are treated as fully populated.
    The count query groups by hostname(). An empty staging table produces no row for that replica, and recordCountWasReached remains true. Therefore, if only one replica sees the inserted IDs
    while others still see empty tables, the method returns success. This recreates the exact race the PR intends to fix. The query must force one count row per expected replica and verify that
    every replica responded with the expected count. I reproduced the zero-row behavior using local ClickHouse.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
```diff
@@ -83,78 +88,238 @@ public void addIds(Collection<? extends Number> ids) {
...
+                                recordCountOnReplica,
+                                pendingIds.size()));
+                        recordCountWasReached = false;
+                        break;
+                    }
+                }
+                if (recordCountWasReached) {
+                    return true;
+                }
+                Thread.sleep(1000 * WAIT_CYCLE_PERIOD_SECONDS);
```
forus: why we sleep here?

forus (approved):

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: Thanks @forus ... I think I understand the issue you are raising. I'll think more deeply. I guess we either would need to query about the cluster composition (to know how many replicas are part of the cluster) and verify the rowcount as well, or we need to try another approach.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: The reason why is connected to this comment from before:
https://github.com/cBioPortal/cbioportal-core/pull/166#issuecomment-4971999773

During my debugging efforts I came to believe that it was risky to use the same connection for several rapid operations, and that it could be causing exceptions to be thrown. When re-establishing the connection from the dbcp connection pool, the connection is first validated to be in a good state before it is made available, so that helps protect against driving a connection into a bad processing state. (that was my theory anyway)

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: I believe a new session is started with each new connection. (Actually that is a good question --- when using a connection pool I wonder if altering settings with `SET setting = value` statements will be passed on to any new calls which obtain that particular connection from the pool or not ... hopefully not)
Anyway, I have tried to adopt a "one statement per connection" standard here, so the mutations_sync setting might need to be re-applied every time anyway.

And a side comment : this is yet a third way of trying to enforce the "wait until mutations propagate through the whole cluster" rule. It would be better if we could have a single approach to that need which is used in all three places (and maybe elsewhere). I will think more.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: This is the pause in between the "try .. retry" cycles while waiting.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: fixed

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: done ; and I also standardized the check that all tables were created into the same pattern

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: fixed

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: I searched for the semantics of the field "is_done" in system.mutations --- it was stated that this occurs once the altered parts have been written to disk. But in a clustered deployment there may be other replicas which have not yet updated their metadata so that they have incorporated the changes which were written to disk (these replicas may still be using old/obsolete parts). That is my understanding. It may not take very much time for the metadata to propagate to other replicas. It also might be true that the metadata transmission also must occur before "is_done" becomes true, but that was not returned from my searches. So in my mind there is still a risk that we are not yet ready to use the new data state which was written by another replica. Therefore I am keeping the check across all replicas to see that the expected conditions are present.

I have added a check to address the issue raised by @forus --- now when we run any check involving clusterAllReplicas() we compare the results against a list of *all replica nodes* that exist in the cluster. If any nodes exist which are not reported in the evaluation of clusterAllReplicas() we continue the wait / retry loop.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: I have replaced this approach of waiting for tables to be created with an approach which mirrors the other wait states --- I now have added a predicate function which checks the metadata in the 'system.tables' table to see whether or not the created table is present/visible. This is within the same try/retry mechanism as the other waits. We now use the same code for waiting for:
1. the creation of the staging_delete_ tables
2. the population of the staging_delete_ tables with the internal identifiers
3. the deletion operation which reads the internal identifiers from the staging_delete_ tables
We wait for (1) to be complete before starting (2)
We wait for (2) to be complete before starting (3)
We wait for (3) to be complete before we try to drop the staging tables.

src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java
sheridancbio: This sleep() call was moved into a finally {} block to make sure it is not skipped should an exception occur in the try{} block.

this addresses the issue raised in this comment:

> [P2] Expected propagation errors exhaust retries immediately. src/main/java/org/mskcc/cbio/portal/dao/ClickHouseBulkDeleter.java:204
> SQL exceptions do not sleep before retrying. A replica returning “table not found” while DDL propagates can consume all seven tolerated exceptions in milliseconds, defeating the intended ten-
> minute wait.

sheridancbio: a test is now done in the finally {} block to see whether another wait cycle will be coming. If not, no sleep is performed.

PR #168, opened 2026-07-31
Use ClickHouseAutoIncrement rather than dao.getLargestId()

sheridancbio:
Several import classes were using their own business logic to increment internal ids based on the highest available integer value in the current database table.

This PR converts that business logic over to using the ClickHouseAutoIncrement class, and will also relocate the assignment of internal identifiers into the dao layer rather than the scripts layer.

### DIFF
--- a/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
+++ b/src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
@@ -650,22 +650,6 @@ public final class DaoMutation {
         return events;
     }
 
-    public static long getLargestMutationEventId() throws DaoException {
-        Connection con = null;
-        PreparedStatement pstmt = null;
-        ResultSet rs = null;
-        try {
-            con = JdbcUtil.getDbConnection(DaoMutation.class);
-            pstmt = con.prepareStatement("SELECT max(`mutation_event_id`) FROM `mutation_event`");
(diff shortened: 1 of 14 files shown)


forus (approved):
Looks good!

src/main/java/org/mskcc/cbio/portal/dao/DaoMutation.java
```diff
@@ -74,70 +74,77 @@ public final class DaoMutation {
     private static final String DELETE_ALTERATION_DRIVER_ANNOTATION = "DELETE from alteration_driver_annotation WHERE genetic_profile_id=? and sample_id=?";
     private static final String DELETE_MUTATION = "DELETE from mutation WHERE genetic_profile_id=? and sample_id=?";
 
-    public static int addMutation(ExtendedMutation mutation, boolean newMutationEvent) throws DaoException {
+    public static void addMutation(ExtendedMutation mutation) throws DaoException {
```
jamesqo: I don't see `ClickHouseAutoIncrement` called in this method? And don't think that `insertRecord` increments the database ID, either.

jamesqo (commented):
Just one comment I have, then you should be good to merge.

Also, I think the fix-delete-race-condition has been merged into master. So if you wanted to rebase off of master before merging that might be a good idea.

PR #170, opened 2026-08-12
Fix ClickHouse read-after-write consistency issue in genetic profile cache

jamesqo:


src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticProfile.java
```diff
@@ -157,7 +157,11 @@ public static int addGeneticProfile(GeneticProfile profile) throws DaoException
             JdbcUtil.closeAll(DaoGeneticProfile.class, con, pstmt, rs);
         }
         
-        reCache();
```
sheridancbio: reCaching between each datatype is also a big hit to study import time. We should rethink this design from scratch soon. I plan to drop non-bulk mode operation completely. Perhaps maintaining a RAM cache to reduce latency is no longer necessary at all.

src/main/java/org/mskcc/cbio/portal/dao/DaoGeneticProfile.java
```diff
@@ -157,7 +157,11 @@ public static int addGeneticProfile(GeneticProfile profile) throws DaoException
...
+        // Add new profile to the local cache directly rather than running a
+        // full reCache().  On ClickHouse Cloud (SharedMergeTree) a SELECT
+        // that follows an INSERT on a different connection may return a stale
+        // snapshot that does not yet contain the freshly-inserted row.
```
sheridancbio: This is a good observation. There is a race condition on rebuilding the cache after making inserts. In theory, inserts are synchronous by default in Clickhouse. But we did see with the bulk deleter that our insert of values into the staging table to hold the list of internal identifiers which need to be deleted *did not completely take effect* when we subsequently used that staging table to perform the deletion. So either we are actually not getting synchronous operation when attaching a binary input stream to the insert operation (our binary stream of TSV values), or there is some delay between the writing of the data parts to disk and the "use" of the updated dataparts on all replicas in a cluster (which I think of as a "distribution of metadata" race condition).

src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
```diff
@@ -148,13 +148,31 @@ public static GeneticProfile loadGeneticProfile(File file) throws IOException, D
...
             // Set `REFERRING_GENETIC_PROFILE_ID`
-            int geneticProfileId = DaoGeneticProfile.getGeneticProfileByStableId(geneticProfile.getStableId()).getGeneticProfileId();
-            geneticProfileLink.setReferringGeneticProfileId(geneticProfileId);
+            GeneticProfile gpLink = DaoGeneticProfile.getGeneticProfileByStableId(geneticProfile.getStableId());
+            if (gpLink == null) {
+                // Fallback: concurrent reCache() may have cleared the cache since addGeneticProfile()
+                DaoGeneticProfile.reCache();
+                gpLink = DaoGeneticProfile.getGeneticProfileByStableId(geneticProfile.getStableId());
+            }
+            if (gpLink == null) {
```
sheridancbio: There has always been a logical problem with the links between genetic profiles. If I am reading this right, it means that when we are importing genetic profiles which refer to each other, we must always import the referred-to genetic profiles before we import the referring-to genetic profiles ... that way we have the internal identifier for the referred-to genetic profile cached before we create the referring-to profile. But do we actually have logic at the scripts layer to insure that we order the genetic profiles correctly? (I don't know). And the references must be organized in a non-circular way .. so if profile A refers to profile B, and profile B refers to profile C, and profile C refers to profile A, then no ordering will work correctly here. Do we detect problems like that in the validator / scripts layer? I doubt it.
I have personally needed to manually delete the (as far as I know) single instance of a study with references between profiles from the database because it has an A->B and B->A cyclical dependency.

We should make a card to think this through and handle references properly / appropriately during all database update operations.

src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
```diff
@@ -148,13 +148,31 @@ public static GeneticProfile loadGeneticProfile(File file) throws IOException, D
...
+                throw new DaoException("Genetic profile could not be found for link after insertion: "
+                        + geneticProfile.getStableId());
+            }
+            geneticProfileLink.setReferringGeneticProfileId(gpLink.getGeneticProfileId());
             DaoGeneticProfileLink.addGeneticProfileLink(geneticProfileLink);
         }
 
         // Get ID
         GeneticProfile gp = DaoGeneticProfile.getGeneticProfileByStableId(geneticProfile.getStableId());
+        if (gp == null) {
```
sheridancbio: It seems to me like this should never happen ... because above we directly inserted the profile into the caches. Having this fallback seems to mean we don't really understand why the faster approach fails sometimes. If we are going to prefer a direct manipulation of the cache rather than a full recache, perhaps we should push to discover why it may be failing.

Having said that ... maybe we are going to be dropping most of the RAM caching here anyway, so maybe it doesn't matter.

src/main/java/org/mskcc/cbio/portal/util/GeneticProfileReader.java
```diff
@@ -168,6 +186,17 @@ private static GeneticProfileLink createGeneticProfileLink(GeneticProfile geneti
             throw new RuntimeException("'source_stable_id' is required in meta file for " + geneticProfile.getStableId());
         }
         GeneticProfile referredGeneticProfile = DaoGeneticProfile.getGeneticProfileByStableId(referredGeneticProfileStableId);
+        if (referredGeneticProfile == null) {
```
sheridancbio: similar to my last comment : a fallback here seems to be sweeping under the rug the fact that we don't understand why our direct manipulation of the cache fails sometimes. If it totally fails sometimes, maybe it partially fails sometimes? If possible we should try to get to the bottom of this. But maybe we should simply move away from RAM caches when we stop supporting non-bulk operations.

sheridancbio (approved):
Approving with a little discomfort. If we think a direct manipulation of the RAM cache of genetic profiles is the way to speed things up (which is fine) it is worrisome to think that sometimes our direct manipulation can fail and we need a fallback to full recaching. If we don't know why it is failing we cannot be sure we "did it right".

But if the data results are equivalent (the data loaded into the database) matches with the prior implementation, then this speedup is ok. I am thinking we can reduce our reliance on these RAM caches of identifiers when we change to doing all-bulk-mode operation. (no longer supporting single record insertions).

PR #175, opened 2026-08-24
clone derived tables as well

sheridancbio:
- change from recreating derived tables (clickhouse.sql) to repopulating derived tables (populate_derived_tables.sql)
- clone the derived tables (schema only) when cloning a database

### DIFF
--- a/scripts/clickhouse_import_support/clone_clickhouse_database.sh
+++ b/scripts/clickhouse_import_support/clone_clickhouse_database.sh
@@ -238,23 +238,30 @@ function clone_all_source_database_tables_to_destination_database() {
 
     while [ "$pos" -lt "$num_tables" ] ; do
         local table_name="${database_table_list[$pos]}"
-        if [ "$table_name" == "$AUTOINCREMENT_SEQUENCE_STATE_TABLENAME" ] || [[ "$table_name" == *_derived ]] ; then
+        if [ "$table_name" == "$AUTOINCREMENT_SEQUENCE_STATE_TABLENAME" ] ; then
             echo "skipping $table_name"
             pos=$(($pos+1))
             continue
+        fi
+        if [[ "$table_name" == *_derived ]] ; then
+            local clone_schema_only="true"
+        else
+            local clone_schema_only="false"
         fi
         echo "cloning $table_name"
         if ! create_destination_database_table_schema_only "$table_name" ; then
             echo "Error : could not create database table schema for $table_name in destination database" >&2
             return 1
         fi
-        if ! copy_source_database_table_data_to_destination "$table_name" ; then
-            echo "Error : could not copy data from table $table_name into destination database" >&2
-            return 1
-        fi
-        if ! destination_table_matches_source_table "$table_name" ; then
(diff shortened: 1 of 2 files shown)


scripts/clickhouse_import_support/download_clickhouse_sql_scripts_py3.py
```diff
@@ -85,7 +85,7 @@ def download_files_from_github(args):
     if not directory_content_response:
         sys.exit(1)
     directory_content = json.loads(directory_content_response.read().decode("utf-8"))
-    files_to_be_downloaded = [x for x in directory_content if os.path.basename(urlparse(x["url"]).path).casefold().endswith(".sql")]
+    files_to_be_downloaded = [x for x in directory_content if os.path.basename(urlparse(x["url"]).path).casefold().endswith("populate_derived_tables.sql")]
```
dippindots: We chanted the filter from `.endswith(".sql")` to `.endswith("populate_derived_tables.sql")`, it looks good and correct to adapt our new changes.

The docstring above (`'Downloads all files ending with \'.sql\'...'`) is now inaccurate and should be updated.

Should we also consider using an exact filename match instead?

scripts/clickhouse_import_support/clone_clickhouse_database.sh
```diff
@@ -235,26 +235,34 @@ function clone_all_source_database_tables_to_destination_database() {
     local pos=0
     local num_tables=${#database_table_list[@]}
     local AUTOINCREMENT_SEQUENCE_STATE_TABLENAME="cbioportal_sequence_state"
+    local clone_schema_only="false"
```
dippindots: `clone_schema_only` is initialized before the loop and correctly reassigned in both branches of the `if/else` on each iteration, so there’s no bug today.

As a small robustness/readability improvement, consider declaring it as a local variable inside the loop, right where it’s set, rather than before the `while`. This would prevent it from accidentally retaining stale state if a future change adds a `continue` between the reset and its use.

dippindots (approved):
LGTM, just had two minor comments, not blocking. Feel free to merge

scripts/clickhouse_import_support/clone_clickhouse_database.sh
sheridancbio: changed as recommended

scripts/clickhouse_import_support/download_clickhouse_sql_scripts_py3.py
sheridancbio: I've updated the docstring to be accurate. I've adjusted the filename comparison from .endswith() to an exact filename match. However, I added an optional command line argument called "--source_filename" to allow a user to specify their own filename to download, or they can also provide a value "*" to cause the program to download *all* files which end in ".sql". By default the program will download only "populate_derived_tables.sql".

dippindots: sounds good, thanks!

sheridancbio:
I've made requested changes and also validated that a successful download was completed before returning a SUCCESS exit status.

